PR #31, opened 2019-06-27
Atom: RichText

### LINKED ISSUE #8: Create RichText Atom
To render html from strings with html code. i.e. richText fields in ContentFull

AndyEPhipps:
Type: patch

Fixes #8


src/components/Atoms/RichText/RichText.js
```diff
@@ -0,0 +1,33 @@
...
+    <RichTextInner
+      align={align}
+      dangerouslySetInnerHTML={{ __html: sanitisedMarkup }}
+      {...rest}
+    />
+  );
+};
+
+RichText.propTypes = {
+  align: PropTypes.string,
```
gusliedke: here we could use `oneOf`
```
.oneOf([
    'left',
    'center',
    'right'
  ]),
```

src/components/Atoms/RichText/RichText.js
```diff
@@ -0,0 +1,33 @@
+import React from 'react';
+import styled from 'styled-components';
+import PropTypes from 'prop-types';
+
+import DOMPurify from 'dompurify';
+
+export const RichTextInner = styled.div`
```
gusliedke: RichTextInner could just be Inner or Wrapper .. In styled component we don't need to worry about naming structure .. we can use Wrapper, Inner, Container etc in all components . it will generate random classes

src/components/Atoms/RichText/RichText.js
AndyEPhipps: ahhh nice, I didn't know you could predefine props like that

src/components/Atoms/RichText/RichText.js
AndyEPhipps: Okeydokey, will change to Inner and use that from here onwards for consistency 👍

PR #35, opened 2019-07-02
fix: Remove dompurify not working with gatsby

gusliedke:
Type: fix

Dompurify will not work with gatsby built app. We will have to sanitise the content at the app level.

### DIFF
--- a/src/components/Atoms/RichText/RichText.test.js
+++ b/src/components/Atoms/RichText/RichText.test.js
@@ -3,8 +3,7 @@ import 'jest-styled-components';
 import renderWithTheme from '../../../hoc/shallowWithTheme';
 import RichText from './RichText';
 
-const unsanitisedHTML =
-  "<p>Here's some copy</p><script>alert('bad things')</script><span>More copy</span>";
+const unsanitisedHTML = "<p>Here's some copy</p><span>More copy</span>";
 
 it('It sanitises markup and renders properly', () => {
   const tree = renderWithTheme(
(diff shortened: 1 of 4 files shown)


src/components/Atoms/RichText/RichText.test.js
```diff
@@ -3,8 +3,7 @@ import 'jest-styled-components';
 import renderWithTheme from '../../../hoc/shallowWithTheme';
 import RichText from './RichText';
 
-const unsanitisedHTML =
-  "<p>Here's some copy</p><script>alert('bad things')</script><span>More copy</span>";
+const unsanitisedHTML = "<p>Here's some copy</p><span>More copy</span>";
```
AndyEPhipps: I guess we should update this var name now too, makes is sound like a bad thing that we've not addressed haha

just... `testHTML` or something?

gusliedke: yes, I will rename it 👍

AndyEPhipps: I'm still surprised this doesn't work properly.. is it just _where_ the sanitise function is being called?

gusliedke: I think we need to use something like Browserify to make it to work on client side. Then we would be adding more dependencies. 

from dompurify doc:

> DOMPurify also works server-side with node.js as well as client-side via Browserify or similar translators. Node.js 0.x is not supported; either io.js or Node.js 4.x or newer is required.

Would be nice to keep it. I will create a ticket for us to implement dompurify with gatsby (maybe gatsby already sanitises rich text?!).

PR #43, opened 2019-07-08
feat: Single message copy as children props

### LINKED ISSUE #29: Review SingleMessage
Add extra fields and variations.

Make it render children instead hard coded fields.

gusliedke:
BREAKING CHANGE: Single message uses children as props

Type: feature

Fixes #29

### Variations
- [x] Copy First (Copy left) default Copy right
- [x] Copy Only (no Image)
- [x] Full Width (Full width image) copy first or copy "second"

Screenshot 2019-07-08 at 11 55 19

### DIFF
--- a/src/components/Atoms/Picture/Picture.js
+++ b/src/components/Atoms/Picture/Picture.js
@@ -1,6 +1,3 @@
-/* eslint-disable react/forbid-prop-types */
-/* eslint-disable react/prop-types */
-/* fix theme, breakpoint and object types */
 import React from 'react';
 import styled, { withTheme } from 'styled-components';
 import PropTypes from 'prop-types';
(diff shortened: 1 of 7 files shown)


src/components/Atoms/Picture/Picture.js
```diff
@@ -58,7 +58,7 @@ const Picture = ({ images, alt, theme, width, height, objectFit }) => {
 };
 
 Picture.propTypes = {
-  images: PropTypes.object.isRequired,
+  images: PropTypes.oneOfType([PropTypes.string, PropTypes.object]).isRequired,
```
AndyEPhipps: I thought `object` wasn't allowed in the ESlint rules we're using?

gusliedke: It's not .. there is a ```eslint-disable``` line at the top :) .. let me fix that with shape

gusliedke: ok, it was ignoring due `theme` type .. I've added a shape type for `theme` as well . I think it works.

PR #44, opened 2019-07-09
Pagination component

### LINKED ISSUE #40: Pagination
Screenshot 2019-07-03 at 16 36 11

lebaz20:
Type: feature

Fixes #40


gusliedke:
@lebaz20 looks nice. Can you have this structure please?
```
Pagination
Pagination/List/List.js
Pagination/Item/Item.js
```
Also let's avoid using classes. All elements should be styled via styled components. Any "modify" like `selected`, should go in as props and dealt in the styled component.

gusliedke:
sorry I pressed close instead Comment. : \

lebaz20:
@gusliedke File structure updated and css classes removed.

src/components/Atoms/Pagination/List/List.js
```diff
@@ -0,0 +1,65 @@
...
+  PageComponent,
+  pages,
+  createURL,
+  onSelect,
+  className,
+  pageComponentProps,
+  ...restProps
+}) => {
+  return (
+    <div className={className}>
```
gusliedke: Can we remove this div with the class and have `ul` and `li` as styled components?

lebaz20: ok, i thought it's just css classes, that we want to get-rid of, will work on that now

src/components/Atoms/Pagination/List/List.style.js
```diff
@@ -0,0 +1,14 @@
+import styled from 'styled-components';
+import List from './List';
+
+const StyledList = styled(List)`
+  ul {
```
gusliedke: As comment above this could be styled components `StyledList`(ul) and `ListItems`(li).

Let's use flex-box instead float.

src/components/Atoms/Pagination/List/List.js
lebaz20: Fixed

src/components/Atoms/Pagination/List/List.style.js
lebaz20: Fixed

gusliedke:
Nice, I think we can remove `className` from ```Pagination.js``` now.

lebaz20:
`classname` removed, I missed the last one in Pagination main class, now removed.

PR #47, opened 2019-07-10
Social media component

### LINKED ISSUE #37: Social icons
Screenshot 2019-07-02 at 15 20 08

svg + link

lebaz20:
Type: feature

Fixes #37


gusliedke:
- Can we have this structure please?
```
SocialIcons
SocialIcons/Icons.js (if we need it)
```
- Can we use our own `svg`s so we don't have more dependencies in new packages?
- Can we use styled components, no class.

Sorry I should have given more info beforehand

lebaz20:
Fix comments above:
- Use styled components instead of CSS classes
- Use our own SVGs
- Use flex instead of float
- Update file structure

gusliedke:
Nice, Can you also optimise our `svgs`? [SVGO](https://www.npmjs.com/package/svgo) is very good and it will remove those illustrator comments.

lebaz20:
> Nice, Can you also optimise our `svgs`? [SVGO](https://www.npmjs.com/package/svgo) is very good and it will remove those illustrator comments.

Nice tool, SVGs are now optimized.

PR #69, opened 2019-07-31
Fix/21 more colour variants for links

### LINKED ISSUE #21: Add more colour variations for Link component
And all inline link colours too.

Heleen-cr:
Type: patch

Fixes #21

- [x] add fix in commit message
- add button colour types to theme and use this for links styled as buttons.
- add variations of inline links


src/components/Atoms/Link/Link.js
```diff
@@ -1,47 +1,45 @@
...
   );
 };
 
 Link.propTypes = {
-  /** Inline is link style */
-  inline: PropTypes.bool,
-  wrapper: PropTypes.bool,
-  /** Button color style */
+  /** Link styling: button, standard, etc. See linkStyles in theme. */
+  linktype: PropTypes.string,
```
gusliedke: We can define the values `PropTypes.oneOf(['button', 'standard'])`

gusliedke: I also would just call it `type`.

src/components/Atoms/Link/Link.md
```diff
@@ -1,33 +1,61 @@
...
-    >
-      <a href="#">Link as button</a>
+    <Link href="#anchor" target="blank" linktype="standard">
+      Link standard
     </Link>
   </div>
   <div style={{ display: 'block', marginBottom: '10px' }}>
-    <Link color="red" href="/test" target="self">
-      Link as button
+    <Link href="#anchor" linktype="bold">
```
gusliedke: Should this be `weight` instead type.

gusliedke: weight could also be done via `Text`.

Heleen-cr: yeah that's true. It used to be one of the link types we had and if you'd use the bold class you'd get a bold link you see. I'll remove it and just keep the standard and standard_white link styles 👍

src/components/Atoms/Link/Link.js
Heleen-cr: nice one, will add that

Heleen-cr:
@gusliedke completely forgot to add the transparent button there, will add now.

src/theme/crTheme/linkStyles.js
```diff
@@ -19,6 +19,15 @@ const linkStyles = {
...
+  bold: {
+    color: color('black'),
+    border: `2px solid ${color('black')}`,
+    weight: '700',
+    padding: '0 2px 1px',
+    hoverColor: color('black'),
+    hoverBorder: `2px solid ${color('black')}`,
+    hoverWeight: '800'
   }
 };
```
LouiaR: Can we used spacing variant on padding

Heleen-cr: like a variable? Not sure what you mean by spacing variant

LouiaR: Yep, you can check an example on this file `DoubleCopy.style.js`

gusliedke: Yeah, it would be nice for us to start using our `spancing` util. ie.`margin-bottom: ${spacing('sm')};` whenever is possible .. margins between elements is a good case for it. Maybe a bit difficult to apply in button padding for instance.. see where it fits best having in mind we need to stay close to current design.

Heleen-cr: ahh ok I see.  I'll have a look to see if it's useful for the link padding. 👍

Heleen-cr: myeah so I think it's not useful for the link padding there. Do you guys want me to introduce a margin-bottom with the spacing variable just in case we need to have multiple buttons on top of each other or something?

gusliedke: I don't think so, margins around buttons and link will be defined by its "parent" the component that will be using them and sorting their position etc. Ready to merge! :)

PR #80, opened 2019-08-03
fix: #79 - window not defined

### LINKED ISSUE #79: Window is not defined in Gatsby build
Window is not defined in Gatsby build

s-barrah:
Type: patch

Fixes #79 

- Fixes `ReferenceError: window is not defined` during Gatsby build as components are rendered server-side where `window` isn't defined. 
- Solved by accessing `window` after component has mounted or checking if `window` is not `undefined` before accessing it.

### DIFF
--- a/src/components/Molecules/Header/Nav/Nav.js
+++ b/src/components/Molecules/Header/Nav/Nav.js
@@ -84,7 +84,7 @@ const MainNav = ({ navItems }) => {
               index={index}
               isSubMenuOpen={!!isSubMenuOpen[group.id]}
             >
-              {!mobile ? (
+              {!isMobile ? (
                 <NavLink
                   href={group.url}
                   inline
--- a/src/components/Molecules/Header/Nav/Nav.js
+++ b/src/components/Molecules/Header/Nav/Nav.js
@@ -22,7 +22,7 @@ const MainNav = ({ navItems }) => {
   const [isSubMenuOpen, setIsSubMenuOpen] = useState({});
   const [isKeyPressed, setIsKeyPressed] = useState({});
   let [width] = useState(null);
-  let [mobile] = useState(null);
+  let mobile;
 
   useEffect(() => {
     // Detect window screen size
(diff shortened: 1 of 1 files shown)


src/components/Molecules/Header/Nav/Nav.js
```diff
@@ -21,16 +21,20 @@ const MainNav = ({ navItems }) => {
...
     event.preventDefault();
     setIsExpandable(!isExpandable);
   };
 
-  // Detect windown screen size
-  const width = window.innerWidth;
-  const mobile = width < sizes.medium;
-
   const toggleSubMenu = item => event => {
     if (mobile) {
```
LouiaR: Nice job @s-barrah, currently toggleBurgerMenu is not working because mobile is set to null

src/components/Molecules/Header/Nav/Nav.js
```diff
@@ -21,16 +21,20 @@ const MainNav = ({ navItems }) => {
   const [isExpandable, setIsExpandable] = useState(false);
   const [isSubMenuOpen, setIsSubMenuOpen] = useState({});
   const [isKeyPressed, setIsKeyPressed] = useState({});
+  let [width] = useState(null);
+  let mobile = useState(false);
+
+  useEffect(() => {
+    // Detect window screen size
+    width = window.innerWidth;
```
gusliedke: wouldn't something like ```width= typeof window !== 'undefined' && window.innerWidth;``` work?

src/components/Molecules/Header/Nav/Nav.js
```diff
@@ -21,16 +21,20 @@ const MainNav = ({ navItems }) => {
   const [isExpandable, setIsExpandable] = useState(false);
   const [isSubMenuOpen, setIsSubMenuOpen] = useState({});
   const [isKeyPressed, setIsKeyPressed] = useState({});
+  let [width] = useState(null);
```
gusliedke: Can we have same as above and better const naming?
```
const [initialWidth, setInitialWidth] = useState(null) // probably we need to set value here for SSR
const [isMobile, setIsMobile] = useState(false);
```

gusliedke:
@LouiaR can you check again if `Header` is still functioning properly with this fix?

LouiaR:
On mobile, when you click to one of the li first level instead to toggle, it redirects you to a new page

gusliedke:
Could you pair with @s-barrah so he knows how to check nav functionality. Maybe a good idea to add some tests checking those toggles.

PR #82, opened 2019-08-05
Header: hook fix warning

### LINKED ISSUE #81: Header: hook fix dependencing
- Fix: 
        . `React Hook useEffect contains a call to 'setIsMobile'. Without a list of dependencies, this can lead to an infinite chain of updates`
        .   `When the submenu is opened on the small screen, if it's not closed when resizing to large it will remain open `

LouiaR:
Type: patch

Fixes #81

### DIFF
--- a/src/components/Molecules/Header/Nav/Nav.style.js
+++ b/src/components/Molecules/Header/Nav/Nav.style.js
@@ -58,7 +58,7 @@ const SubNavMenu = styled.ul`
   background-color: ${({ theme }) => theme.color('deep_violet')};
 
   @media ${({ theme }) => theme.breakpoint('medium')} {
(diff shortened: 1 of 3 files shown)


src/components/Molecules/Header/Nav/Nav.style.js
```diff
@@ -58,7 +58,7 @@ const SubNavMenu = styled.ul`
   background-color: ${({ theme }) => theme.color('deep_violet')};
 
   @media ${({ theme }) => theme.breakpoint('medium')} {
-    display: ${({ isSubMenuOpen }) => (isSubMenuOpen ? 'none' : 'none')};
+    display: none;
```
gusliedke: Why do we need this if we've already got `display` below?

PR #84, opened 2019-08-06
feat: Add whitelisted util

### LINKED ISSUE #83: Create whiteListed util
Function to check if `url` is white listed.

gusliedke:
Type: feat

Fixes #83


gusliedke:
@AndyEPhipps yeah I think it make more sense to keep `whitelist` here so it can be used by ```Navs```

AndyEPhipps:
@gusliedke coolcool; we can always make it CMSable in the future but it's probably over-engineering the problem

PR #88, opened 2019-08-08
Use exiting SocialIcons comp in Footer

### LINKED ISSUE #87: Use existing Socialicons comp in Footer

AndyEPhipps:
Fixes #87


src/components/Molecules/Footer/Footer.style.js
```diff
@@ -48,4 +48,34 @@ const FooterCopyright = styled.div`
   }
 `;
 
-export { FooterWrapper, InnerWrapper, FooterCopyright, FooterBranding };
+const SocialIconWrapper = styled.div`
+  ul {
```
LouiaR: Hey @AndyEPhipps , instead of nesting ul here could you extend SocialIcons atom styling instead (not really issue)

gusliedke: Good idea!

AndyEPhipps: Yeah, I had tried that already, but the way these SocialIcons have been set up, the component doesn't represent the _UL_ itself:

```
const FooterSocialIcons = styled(SocialIcons)`
  background: green;
`;
```

```
const Footer = ({ navItems, footerCopy, campaign, ...rest }) => {
  return (
    <FooterWrapper navItems {...rest}>
      <InnerWrapper>
        <SocialIconWrapper>
          <FooterSocialIcons campaign={campaign} />
        </SocialIconWrapper>
```

Screen Shot 2019-08-09 at 11 02 29

AndyEPhipps: Or am I missing something?

LouiaR: No, you right

AndyEPhipps: I looked over the SocialIcon code and I still can't figure out what it's doing that though, it looks like it SHOULD represent the UL:

```
const SocialIcons = ({ campaign, ...restProps }) => {
  const links = getLinks(campaign);
  return (
    <StyledList>
      {Object.keys(icons).map(brand => (
        <StyledItem key={brand}>
          <Icon
            icon={icons[brand]}
            href={links[brand]}
            brand={brand}
            {...restProps}
          />
        </StyledItem>
      ))}
    </StyledList>
  );
};
```

gusliedke:
```
const StyledList = styled.ul`
  display: flex;
  list-style-type: none;
`;

const StyledItem = styled.li`
  width: 80px;
  padding: 10px;
`;

const StyledLink = styled.a`
  text-decoration: none;
  cursor: pointer;
`;

const StyledImage = styled.img`
  width: 100%;
`;
```
We should be changing those styled components not overriding it. Those styled component are in `socialIcons.js` and `icons.js`.

AndyEPhipps: Just had a chat with Gus, figured I'd echo it here for @LouiaR (and @Heleen-cr );

Since this `SocialLink` is currently ONLY being used in the `Footer`, it makes sense to put all styles within the `SocialLink` itself, and any new uses in the future will be overridden within that parent component context.

The reason I couldn't target SocialIcon's `StyledList`, `StyledItem` etc. within my parent `Footer` component was because those elements weren't being exported as part of `SocialLink`. Once they're exported, you can target them with CSS, and achieve that 'future overriding' I mentioned.

You probably knew all this already but I didn't! Thanks guys 👍

Heleen-cr: yeah ok that makes sense

PR #107, opened 2019-08-14
Fix: update Footer and Header menus to reflect how Contentful produces menus

### LINKED ISSUE #106: Update Footer data object
Just need to rename a few fields to match Contentful

AndyEPhipps:
Type: patch

Fixes #106


AndyEPhipps:
DON'T MERGE FOR NOW! I'll need to make some adjustments within comicrelief-contentful first before this goes anywhere.

Will see if I can change my 'fix:' commit message to a 'breaking change:' one, to be on the safe side.

PR #134, opened 2019-08-30
fix: fix card image

LouiaR:
Type: fix

Issue: the image  is not displayed on Card component

Screen Shot 2019-08-30 at 14 10 37

### DIFF
--- a/src/components/Molecules/Card/Card.md
+++ b/src/components/Molecules/Card/Card.md
@@ -1,11 +1,11 @@
-Card
+# Card
 
 ```js
 const data = require('../../../styleguide/data/data').default;
 import Text from '../../Atoms/Text/Text';
 import Link from '../../Atoms/Link/Link';
 
-<Card image={data.image} images={data.image} backgroundColor="yellow">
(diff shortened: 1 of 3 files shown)


src/components/Molecules/Card/Card.md
```diff
@@ -1,11 +1,11 @@
-Card
+# Card
 
 ```js
 const data = require('../../../styleguide/data/data').default;
 import Text from '../../Atoms/Text/Text';
 import Link from '../../Atoms/Link/Link';
 
-<Card image={data.image} images={data.image} backgroundColor="yellow">
+<Card imageLow={data.image} images={data.image} backgroundColor="yellow">
```
gusliedke: We need add `images` ie. `images={data.images}`

LouiaR: Yeah It's the props name expected by `Picture` atom

gusliedke: I mean in `images={data.image}` you need to add `s` ie `image={data.images}`

LouiaR: Sure

PR #143, opened 2019-09-12
Update article teaser

### LINKED ISSUE #142: Update article teaser
- Remove padding on image
- make image take full height when content increases

LouiaR:
Type: fix

Fixes #142


src/components/Molecules/ArticleTeaser/ArticleTeaser.js
```diff
@@ -31,36 +33,48 @@ const Link = styled(link)`
...
 const CopyWrapper = styled.div`
   padding: ${spacing('l')};
-  @media ${({ theme }) => theme.breakpoint('small')} {
-    width: ${({ category }) => (category ? '100%' : '55%')};
-  }
-
-  @media ${({ theme }) => theme.breakpoint('medium')} {
-    width: 100%;
-  }
+  word-break: break-all;
```
AndyEPhipps: We don't wanna be breaking-up words like this:




Could we remove that rule? Looks like the text is already set-up to break properly

LouiaR: Sure, done

PR #148, opened 2019-09-17
Improve header nav

### LINKED ISSUE #147: Main menu bug
menu_bug

Menu collapses when user try to click child item.

LouiaR:
Type: patch

Fixes #147


src/components/Atoms/Link/Link.style.js
```diff
@@ -3,9 +3,6 @@ import styled, { css } from 'styled-components';
 const buttonStyle = () => css`
   display: inline-block;
   padding: 11px 17px;
```
gusliedke: Can we use `spacing` here?

LouiaR: Sure

PR #155, opened 2019-09-26
Single msg vh fix

gusliedke:
Type: fix


LouiaR:
Hey @gusliedke should we use `flex: 0 0 50%` on  both `Copy` and `Media`  instead of `width`  so that they will always be at the same width instead of their width be dictated by their content. what do you think ?

gusliedke:
> Hey @gusliedke should we use `flex: 0 0 50%` on both `Copy` and `Media` instead of `width` so that they will always be at the same width instead of their width be dictated by their content. what do you think ?

@LouiaR Copy has `width: 50%` this should prevent copy to shrink or grow .. looks like all variations are working as expected .. I like your suggestion as we are using flex we could use flex-basis instead width. But at the same time I don't want to give a chance now to something go wrong.. let's keep an eye and at some point we can refactor as you said.

PR #156, opened 2019-09-27
fix: flex grow

gusliedke:
Type: fix


gusliedke:
sorry @LouiaR it grew in that case you showed me .. can you please check if this change fix that issue?

PR #165, opened 2019-10-10
ESU modal and header icons

### LINKED ISSUE #164: ESU modal, ESU & Search icons
Relates to https://github.com/comicrelief/comicrelief-contentful/issues/187

AndyEPhipps:
Type: patch

Fixes #164


src/components/Molecules/HeaderEsuWithIcon/HeaderEsuWithIcon.js
```diff
@@ -0,0 +1,101 @@
...
+  const renderESU = () => {
+    return (
+      <HeaderESU
+        title={title}
+        topCopy={topCopy}
+        successCopy={successCopy}
+        privacyCopy={privacyCopy}
+        isSuccess={isSuccess}
+        subscribe={() => setIsSuccess(!isSuccess)}
+        errorMsg={isErroring}
```
lebaz20: Can we expose those props instead of handling them internally ?
That way, we can do the request to our backend, and validate inside `comicrelief-contentful` repo.,
or may be do the validation here instead, just like https://github.com/comicrelief/comicrelief-contentful/blob/master/src/components/EmailSignUp/EmailSignUp.js#L4.

package.json
```diff
@@ -65,8 +65,6 @@
   },
   "devDependencies": {
     "babel-eslint": "^10.0.2",
-    "enzyme": "^3.10.0",
-    "enzyme-adapter-react-16": "^1.14.0",
```
lebaz20: Removed those packages, as i can not find any use for them!

src/components/Molecules/HeaderEsuWithIcon/HeaderEsuWithIcon.test.js
```diff
@@ -0,0 +1,55 @@
...
+
+it('renders correctly with modal open', () => {
+  const tree = renderWithTheme(
+    <div>
+      <HeaderEsuWithIcon
+        title={title}
+        topCopy={topCopy}
+        privacyCopy={privacyCopy}
+        successCopy={successCopy}
+        isESUOpen
```
lebaz20: Added a test case for `open modal` using initial value for `isESUOpen` state variable, that's set via a prop with the same name.

AndyEPhipps: Ahhh excellent, totally forgot about tests as i finished this last thing Friday; thank @lebaz20

src/components/Molecules/HeaderEsuWithIcon/HeaderEsuWithIcon.js
AndyEPhipps: Can do! I think this was from before I saw how you were handling the existing ESU within Contentful, will make it work in the same way 👍

PR #166, opened 2019-10-14
feat: Video header component

### LINKED ISSUE #163: Video Header
### Video as background for Headers

Screenshot 2019-10-10 at 12 41 27

gusliedke:
Type: feature

Fixes #163 

video


src/components/Molecules/VideoBanner/VideoBanner.md
```diff
@@ -0,0 +1,8 @@
+```js
+const image =
+  'https://www.comicrelief.com/sites/default/files/2019-10/Creativists_Logo_Colour_on%20background.png';
```
AndyEPhipps: Can we get these files into the repo?

src/components/Molecules/VideoBanner/VideoBanner.js
```diff
@@ -0,0 +1,38 @@
...
+  const onPlay = () => {
+    videoEl.current.play();
+  };
+
+  useEffect(() => {
+    onPlay();
+  });
+
+  return (
+    <Video ref={videoEl} autoPlay={autoPlay} loop={loop} muted playsinline>
```
AndyEPhipps: I guess it doesn't _really_ matter here (as we can "hard set" it within the Contentful integration), but we probably need to ensure these bg videos **don't** loop; it was part of an accessibility update following a DAC audit and we don't wanna give CMS options to something that'll have a negative impact.

AndyEPhipps: Also, don't we need to pass the Poster attr?

src/components/Molecules/VideoBanner/VideoBanner.js
```diff
@@ -0,0 +1,38 @@
...
+      <source src={video} type="video/mp4" />
+      Your browser does not support video.
+    </Video>
+  );
+};
+
+VideoBanner.propTypes = {
+  video: PropTypes.string.isRequired,
+  autoPlay: PropTypes.bool.isRequired,
+  loop: PropTypes.bool.isRequired
```
AndyEPhipps: Do we need a Poster prop here?

src/components/Molecules/VideoBanner/VideoBanner.js
gusliedke: Yes, height we can be defined in gatsby app. Ok I will make sure loop is always false.

Adding back poster `attr`! I tried to use bg image instead because poster wasn't filling the container. I want to test it in real use to see how it will look.

src/components/Molecules/VideoBanner/VideoBanner.md
```diff
@@ -0,0 +1,7 @@
+```js
+const poster = require('../../../styleguide/assets/poster.png');
+const src =
+  'https://www.comicrelief.com/sites/default/files/downloads/Creativists_Logo_Web_small_V2_0.mp4';
```
AndyEPhipps: Are we able to add this to the repo too?

gusliedke: it's 2.4mb .. let's not save it locally at this point

PR #180, opened 2019-10-29
fix img res ie 11

gusliedke:
Type: fix

Fixes: https://github.com/comicrelief/comicrelief-contentful/issues/238


gusliedke:
Passed QA https://github.com/comicrelief/comicrelief-contentful/issues/238#issuecomment-548845491

PR #181, opened 2019-10-29
Single message 50:50

### LINKED ISSUE #179: Single msg 50:50
Single message row image is not 50:50 in width on IE11.

_Use `width` instead `flex-basis`_

LouiaR:
Type: patch

Fixes #179


gusliedke:
Styleguidist doesn't work on IE11. We need to provide an env in gatsby with this change for QA to test it.

PR #187, opened 2019-11-04
Membership update: fix form's width auto scaling  with content

### LINKED ISSUE #186: Update membership: fix form'width
On ie form's width scales according to content length

Screen Shot 2019-11-04 at 12 43 29

LouiaR:
Type: patch

Fixes #186

### DIFF
--- a/src/components/Molecules/Membership/Form/Form.js
+++ b/src/components/Molecules/Membership/Form/Form.js
@@ -178,7 +178,7 @@ const Signup = ({
         </FormFieldset>
         <Button type="submit">Donate</Button>
         {errorMsg && (
-          <Error as="p">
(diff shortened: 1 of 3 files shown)


src/components/Molecules/Membership/Form/Form.js
```diff
@@ -178,7 +178,7 @@ const Signup = ({
         </FormFieldset>
         <Button type="submit">Donate</Button>
         {errorMsg && (
-          <Error>
+          <Error as="p">
```
gusliedke: You can use `tag="p"` Error is Text component.

LouiaR: Done

PR #197, opened 2019-11-11
single message copy only spacing udpate

Heleen-cr:
preview single message h1 only (title row) here: https://headless-pr.comicrelief.com/287-single-msg-h1-update/contact-us/

compare with title row on a default page type: https://headless-pr.comicrelief.com/287-single-msg-h1-update/terms-of-use/

### DIFF
--- a/src/components/Molecules/SingleMessage/SingleMessage.style.js
+++ b/src/components/Molecules/SingleMessage/SingleMessage.style.js
@@ -48,7 +48,6 @@ const Copy = styled.div`
     justify-content: center;
     padding: ${spacing('l')} ${spacing('xl')};
   }
-
   ${props =>
     props.fullImage &&
     css`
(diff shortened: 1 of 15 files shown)


src/components/Molecules/SingleMessage/SingleMessage.style.js
```diff
@@ -79,7 +78,18 @@ const Copy = styled.div`
           }
           margin: auto;
           padding: 100px 20px;
```
gusliedke: Can we use `spacing` here? Maybe try "m" for mobile.

src/components/Molecules/SingleMessage/SingleMessage.style.js
```diff
@@ -79,7 +78,18 @@ const Copy = styled.div`
...
-        `}
+          // move empty p tag up with negative margin equal to h1's margin
+          & p:empty {
```
gusliedke: I just learnt about `:empty` selector :) .. Can we try to check if it's empty not null in **RichText** component?
```
[BLOCKS.PARAGRAPH]: (node, children) => {
      const fontSize = options.type === 'card' ? 'sm' : 'l';
      **if node !== ""**
      return (
        <Text tag="p" color={changeColor(options.background)} size={fontSize}>
          {children}
        </Text>
      **return null**
      );
    },
```

PR #202, opened 2019-11-15
feat: Add schools form dropdown option

gusliedke:
Type: patch

Fixes #

- Add dropdown step on Schools form


gusliedke:
> Would be useful if we can have test cases for schools ESU i guess, right ?

Sure 👍

PR #205, opened 2019-11-21
single msg - optional body field

Heleen-cr:


gusliedke:
Can you create an example in `md` file? Also add a snapshot test for this variation.

src/components/Molecules/SingleMessage/SingleMessage.test.js
```diff
@@ -68,6 +68,21 @@ it('renders Single Message with full width correctly', () => {
...
+it('renders Single Message with full width image and no text correctly', () => {
+  const tree = renderWithTheme(
+    <SingleMessage
+      backgroundColor="purple"
+      imageSet={data.images}
+      image={data.image}
+      imageAltText=""
+      copyFirst={false}
+      fullImage
+    ></SingleMessage>
```
LouiaR: I think we don't need to add `</SingleMessage>`

src/components/Molecules/SingleMessage/SingleMessage.md
```diff
@@ -52,6 +52,24 @@ import Link from '../../Atoms/Link/Link';
...
+
+<SingleMessage
+  backgroundColor="purple_dark"
+  imageSet={data.images}
+  image={data.image}
+  imageAltText=""
+  copyFirst={false}
+  fullImage={true}
+  vhFull
+></SingleMessage>;
```
LouiaR: I think we don't need to add </SingleMessage>

src/components/Molecules/SingleMessage/SingleMessage.test.js
Heleen-cr: indeed we don't

PR #219, opened 2020-01-13
H1 text cutting off on schools fundraising ideas pages on iPads Pro portrait view

LouiaR:
screenshot-live browserstack com-2019 12 23-13_44_55

Issue is here: https://github.com/comicrelief/comicrelief-contentful/issues/448

### DIFF
--- a/src/theme/shared/global.css
+++ b/src/theme/shared/global.css
@@ -1,11 +1,14 @@
 @import url('https://fonts.googleapis.com/css?family=Anton%7CMontserrat:500,700,800&display=swap');
 
 html {
+  -webkit-box-sizing: border-box;
   box-sizing: border-box;
 }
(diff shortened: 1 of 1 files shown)


src/theme/shared/global.css
```diff
@@ -1,11 +1,14 @@
...
 
 html {
+  -webkit-box-sizing: border-box;
   box-sizing: border-box;
 }
+
 *,
 *:before,
 *:after {
+  -webkit-box-sizing: inherit;
```
gusliedke: do we need to add vender prefix?

LouiaR: No really, it added automatically by the editor

LouiaR: Removed

PR #220, opened 2020-01-14
483 Search accessibility issues

lebaz20:
Type: patch

Fixes comicrelief/comicrelief-contentful#483

- Allow extra props like autofocus to be passed to search input
- Use background prop in IE only as mask is not supported there


lebaz20:
Unfortunately commit message will not create new release, but i will need to do other changes probably, so will keep that in mind next mind!

PR #228, opened 2020-01-17
Homepage not linked from logo for screen readers

LouiaR:
Type: patch

Issue  is here: https://github.com/comicrelief/comicrelief-contentful/issues/485

### DIFF
--- a/src/components/Atoms/Logo/Logo.js
+++ b/src/components/Atoms/Logo/Logo.js
@@ -36,8 +36,8 @@ const Logo = ({ rotate, sizeSm, sizeMd, campaign }) => {
         src={campaign === 'comicrelief' ? crLogo : srLogo}
         alt={
           campaign === 'comicrelief'
-            ? 'Comic Relief homepage'
-            : 'Sport Relief homepage'
+            ? 'Comic Relief logo - home page'
+            : 'Sport Relief logo - home page'
         }
         aria-label="logo"
       />
--- a/src/components/Atoms/Logo/Logo.js
+++ b/src/components/Atoms/Logo/Logo.js
@@ -35,11 +35,8 @@ const Logo = ({ rotate, sizeSm, sizeMd, campaign }) => {
       <Image
         src={campaign === 'comicrelief' ? crLogo : srLogo}
         alt={
-          campaign === 'comicrelief'
-            ? 'Comic Relief logo - home page'
-            : 'Sport Relief logo - home page'
+          campaign === 'comicrelief' ? 'Comic Relief logo' : 'Sport Relief logo'
         }
-        aria-label="logo"
       />
     </LogoWrapper>
   );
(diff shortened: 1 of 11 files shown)


src/components/Atoms/Logo/Logo.test.js
```diff
@@ -32,12 +32,11 @@ it('renders correctly', () => {
     }
 
     <div
-      aria-label="logo"
       className="c0"
       rotate={1}
     >
       <img
-        alt="Comic Relief logo"
+        alt="Comic Relief homepage"
```
gusliedke: Please change to `Comic Relief logo – home page`

src/components/Atoms/Logo/Logo.js
```diff
@@ -37,7 +35,9 @@ const Logo = ({ rotate, sizeSm, sizeMd, campaign }) => {
       <Image
         src={campaign === 'comicrelief' ? crLogo : srLogo}
         alt={
-          campaign === 'comicrelief' ? 'Comic Relief logo' : 'Sport Relief logo'
+          campaign === 'comicrelief'
+            ? 'Comic Relief logo - home page'
```
gusliedke: @LouiaR sorry, this was a recommendation from DAC report. But thinking better I don't think it make sense to use home page here.. this logo is also used in the footer. Let's just use `Comic Relief logo` and on CR we add `title="homepage"` to the link and a visually hidden span.

LouiaR: @gusliedke what about add `alt='comicrelief.com'`

src/components/Molecules/Footer/Footer.js
```diff
@@ -10,19 +10,27 @@ import {
...
 } from './Footer.style';
 
 const Footer = ({ navItems, footerCopy, campaign, ...rest }) => {
   return (
     <FooterWrapper navItems {...rest}>
       <InnerWrapper>
         <SocialIconWrapper>
-          <SocialIcons campaign={campaign} />
+          <SocialIcons
+            campaign={campaign
```
gusliedke: Can you create a new const `campaignName` and use `.replace(/ +/g, "").join('').toLowerCase()` instead? So we can read what's doing.

src/components/Molecules/Footer/Footer.js
```diff
@@ -10,19 +10,27 @@ import {
...
   FooterCopyright,
-  SocialIconWrapper
+  SocialIconWrapper,
+  Brand
 } from './Footer.style';
 
 const Footer = ({ navItems, footerCopy, campaign, ...rest }) => {
+  // Concatenate campaign in one string
+  const campaignName = campaign
+    .split(' ')
```
gusliedke: Can you create a new const campaignName and use .replace(/ +/g, "").join('').toLowerCase() instead? So we can read what's doing.

src/components/Molecules/Footer/Footer.js
```diff
@@ -10,19 +10,24 @@ import {
   InnerWrapper,
   FooterBranding,
   FooterCopyright,
-  SocialIconWrapper
+  SocialIconWrapper,
+  Brand
 } from './Footer.style';
 
 const Footer = ({ navItems, footerCopy, campaign, ...rest }) => {
+  // Concatenate campaign in one string
```
gusliedke: Can u just replace to `Remove white scape between words`?

PR #240, opened 2020-01-29
add icons to button

### LINKED ISSUE #233: Iconography on buttons
- Download
- External link
- favorite
- -> next

gusliedke:
Type: feat

Fixes #233 

### icons to button

Add icon as child props as I don't want to add more logic to `Link`. And we can have those icons elsewhere.

icons

Env https://deploy-preview-240--cr-component-library.netlify.com/#!/Link/2


src/components/Atoms/Icons/Download.js
```diff
@@ -0,0 +1,82 @@
...
+    </svg>
+  );
+};
+
+Download.propTypes = {
+  colour: PropTypes.string
+};
+
+Download.defaultProps = {
+  colour: '#ffffff'
```
LouiaR: Can we use color variable here

PR #246, opened 2020-02-06
load player on click

gusliedke:
Type: patch

Fixes video performance

### DIFF
--- a/src/components/Molecules/SingleMessage/SingleMessage.js
+++ b/src/components/Molecules/SingleMessage/SingleMessage.js
@@ -39,6 +39,7 @@ const SingleMessage = ({
   // States to track video status
   const [isInitialised, setIsInitialised] = useState(false);
   const [isPlaying, setIsPlaying] = useState(false);
+  const [onPlay, setonPlay] = useState(false);
   const [isBuffering, setIsBuffering] = useState(false);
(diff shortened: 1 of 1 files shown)


gusliedke:
@AndyEPhipps trying to load youtube player onPlay and setting a time out to give it time to find video ID.

src/components/Molecules/SingleMessage/SingleMessage.js
```diff
@@ -71,32 +72,33 @@ const SingleMessage = ({
...
-  }, [hasVideo, isInitialised, uniqueID, videoID]);
+  }, [hasVideo, isInitialised, uniqueID, videoID, onPlay]);
 
   const handlePlay = thisUniqueID => {
+    setonPlay(true);
     // Trigger play and update video state
     const thisVideoID = `${thisUniqueID}__video`;
 
-    allPlayers[thisVideoID].playVideo();
+    setTimeout(() => {
```
AndyEPhipps: Hmm.. is there not a nicer way of doing this? There's no guarantee that this'll load in time on a slow connection

gusliedke: As per our chat this will load the iframe yt player. Just tested with multi videos and it's working.

PR #254, opened 2020-02-24
Teaser reDesign

### LINKED ISSUE #253: Teaser card

gusliedke:
Type: patch

Fixes #253 

image


src/components/Molecules/ArticleTeaser/ArticleTeaser.js
```diff
@@ -18,6 +18,13 @@ const Wrapper = styled.article`
   height: 100%;
   display: flex;
   background-color: #fff;
```
LouiaR: Can we use the variable colour here

PR #257, opened 2020-02-25
Box component

### LINKED ISSUE #256: Box component
image

gusliedke:
Type: patch

Fixes #256


src/components/Molecules/Box/Box.js
```diff
@@ -0,0 +1,82 @@
...
+  @media ${({ theme }) => theme.breakpoint('small')} {
+    flex-direction: row;
+  }
+
+  @media ${({ theme }) => theme.breakpoint('medium')} {
+    flex-direction: column;
+  }
+`;
+
+const Image = styled.div`
```
LouiaR: Can you rename as it's not img tag

gusliedke: Hi, it doesn't need to describe what the tag is. Same for Copy. Let's keep names simple.. only use ImageWrapper if you already used Image.

PR #279, opened 2020-04-24
Add noopener to nav for external links

gusliedke:
Type: patch

https://deploy-preview-279--cr-component-library.netlify.app/#!/Header/3

image

now we have `haspopup` only for items with submenus and `noopner` for external links

### DIFF
--- a/src/components/Organisms/Header/Nav/Nav.js
+++ b/src/components/Organisms/Header/Nav/Nav.js
@@ -87,8 +87,8 @@ const MainNav = ({ navItems }) => {
             /* Determine which field represents our url path */
             let thisUrl = NavHelper(thisFirstChild);
             const relNoopener = !whiteListed(thisUrl) && 'noopener';
-            const haspopup =
-              group.links && group.links.length > 1 ? 'true' : null;
+            const hasSubMenu = group.links && group.links.length > 1;
+            const hasPopUp = hasSubMenu ? 'true' : null;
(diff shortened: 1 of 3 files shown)


src/components/Organisms/Header/Nav/Nav.js
```diff
@@ -85,7 +86,9 @@ const MainNav = ({ navItems }) => {
 
             /* Determine which field represents our url path */
             let thisUrl = NavHelper(thisFirstChild);
-
+            const relNoopener = !whiteListed(thisUrl) && 'noopener';
+            const hasSubmenu =
+              group.links && group.links.length > 1 ? 'true' : null;
```
AndyEPhipps, line 91: Can we not use true boolean values here? 

A `"true"` string and `null` seem a bit weird..

gusliedke: Just checked. I can't use it for submenu logic. I will change it 👍 

I need this way for `haspopup` to ignore the attribute if null but keep it and add string "true".

AndyEPhipps: _Personally_, I'd keep that `hasSubMenu` boolean logic up top (nicer to see the vars together beforehand), and just made a second var for `hasPopUp`, based on `hasSubMenu`:

`hasPopUp = hasSubMenu ? 'true' : null
`

PR #287, opened 2020-05-20
Fix: Issue with Input component duplicating props & adding event handlers twice

### LINKED ISSUE #290: Fix issue with Input component duplicating props (event handlers in particular)
The `<Input>` component was passing additional props captured by `...rest` to two of its child components, Label and InputField (with a few exceptions which were prevented being added to Label)

This duplication was a particular problem for event handlers - e.g. any onChange or onClick prop provided to Input would fire twice.

Classes inserted by styled components were also being added to both children, rather than just the outer element.

tomisaacmarshall:
Fixes #290 

Type: patch

- Alters the Input component so that props captured by `...rest` (including event handlers) are only added to the InputField element
- Adds className prop to Input and passes this to the outer element, Label, only (previously className was being added to both elements)

### DIFF
--- a/src/components/Atoms/Input/Input.js
+++ b/src/components/Atoms/Input/Input.js
@@ -81,13 +81,14 @@ const Input = ({
   showLabel,
   type,
   hasAria,
+  className,
   labelProps,
   ...rest
 }) => {
   const error = errorMsg && errorMsg.length > 0;
 
   return (
-    <Label htmlFor={id} {...labelProps}>
+    <Label htmlFor={id} className={className} {...labelProps}>
       <TextLabel showLabel={showLabel} weight="bold">
         {label}
       </TextLabel>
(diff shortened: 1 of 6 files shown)


adamclark-dev:
@AndyEPhipps @LouiaR @gusliedke could you review

AndyEPhipps:
Makes sense to me, @tomisaacmarshall! 

In order to update the snapshot (so your Circle CI build will pass), you need to run `yarn test -u` locally.

Although it's a pretty minor change, we'd normally pull in this PR version of Component Library into an PR env of the context it's being used in (CR.com in this instance) so make sure there's been unexpected side effects _**in-situ**_

AndyEPhipps:
Also (although it's not documented), but you wanna use a fix:/feat:/BREAKING CHANGE: prefix to one of the commit messages.

It's used by Semantic Versioning to determine which version to bump the package to with this PR.

It's documented here: https://github.com/comicrelief/pattern-lab, will update the docs here shortly.

gusliedke:
Hi Tom, feel free to refactor / clean up Input component. We use it only in Search, Donate widget, EmailSignUp and Membership components. If you change it just update the snap tests, create a PR as per Andy's instructions and we can ask Krupa to test those specifics affected components in CR app.

tomisaacmarshall:
@gusliedke @AndyEPhipps Thanks for comments!

I've been a bit bolder with the changes now, as well as running the snapshot update and adding the commit prefix

Don't fully understand the PR bit though @AndyEPhipps ?

PR #291, opened 2020-05-27
Feat/big night logo

gusliedke:
Type: patch

Add Big night in logo
image


LouiaR:
Not an issue, can you add an example of BN logo.

PR #296, opened 2020-06-03
fix: Display Select component's hint/placeholder

tomisaacmarshall:
Type: patch

The Select component was displaying its first option on initial render.

This PR tweaks the component so that it displays the hint/placeholder until a selection is made.


src/components/Atoms/SelectField/SelectField.js
```diff
@@ -83,13 +83,16 @@ const Select = ({
...
   const error = errorMsg && errorMsg.length > 0;
+
   return (
     <Label>
       <LabelText hideLabel={hideLabel} weight="bold">
         {label}
       </LabelText>
-      <StyledInput {...rest} error={error ? 1 : 0}>
-        <option disabled>{description}</option>
+      <StyledSelect {...rest} error={error ? 1 : 0}>
```
gusliedke, line 92: can't error just be `error={error}` ?

tomisaacmarshall: Yes I would have thought so but this isn't in my changes, I was just renaming the component as StyledInput confused me a little bit. Not entirely sure what the error attribute is, although looks like it might be passing a value to styled components?

PR #298, opened 2020-06-08
Marketing preference

gusliedke:
Type: feat

https://deploy-preview-298--cr-component-library.netlify.app/#!/MarketingPreferences

- [x] style radios
- [x] add validation
- [ ] add test


src/components/Organisms/MarketingPreferences/CheckAnswer.js
```diff
@@ -0,0 +1,36 @@
...
+        value="no"
+        ref={register}
+      />
+      No, thanks
+    </label>
+  </Text>
+);
+
+CheckAnswer.propTypes = {
+  register: PropTypes.objectOf.isRequired,
```
tomisaacmarshall, line 32: I think `objectOf` requires an argument?

src/components/Organisms/MarketingPreferences/InputController.js
```diff
@@ -0,0 +1,29 @@
...
+  };
+
+  return <Controller {...props} />;
+};
+
+InputController.propTypes = {
+  fieldName: PropTypes.string.isRequired,
+  label: PropTypes.string.isRequired,
+  errors: PropTypes.objectOf.isRequired,
+  control: PropTypes.objectOf.isRequired
```
tomisaacmarshall, line 26: Same here with objectOf

src/components/Organisms/MarketingPreferences/MarketingPreferences.js
```diff
@@ -0,0 +1,175 @@
...
+        <Head>
+          <Text tag="h3" size="l" family="Anton" uppercase weight="400">
+            Email
+          </Text>
+          <CheckAnswer name="email" register={register} />
+        </Head>
+        <Body>
+          <Text tag="p" color="grey">
+            You will receive the latest fundraising activities updates and
+            marketing material from Comic Relief brand.
```
tomisaacmarshall, line 30: Is this the actual text or just placeholder? Wording is a little off I think

src/components/Organisms/MarketingPreferences/MarketingPreferences.js
```diff
@@ -0,0 +1,175 @@
...
+          </Text>
+          {watchEmail === 'yes' && (
+            <InputController
+              placeholder="Email Address"
+              fieldName="emailAddress"
```
tomisaacmarshall, line 35: I think all the field names should be prefixed so that they are unlikely to clash with other fields in the form that's using this component

src/components/Organisms/MarketingPreferences/MarketingPreferences.js
gusliedke: I copied it from design. I will make it as child props so it can be editable.

src/components/Organisms/MarketingPreferences/MarketingPreferences.js
```diff
@@ -0,0 +1,175 @@
+import React from 'react';
+import { useFormContext } from 'react-hook-form';
+
+import Text from '../../Atoms/Text/Text';
+import Link from '../../Atoms/Link/Link';
+import InputController from './InputController';
+import CheckAnswer from './CheckAnswer';
+import { Head, Body } from './MarketingPreferences.style';
+
+const MarketingPreferences = () => {
```
tomisaacmarshall, line 10: I think it would be good to be able to have an initial values prop to pre-populate the fields. E.g. on the fundraise form, we already have their email address at this point, so would be good if they don't have to re-enter it

src/components/Organisms/MarketingPreferences/MarketingPreferences.js
```diff
@@ -0,0 +1,175 @@
...
+          href="https://www.comicrelief.com/update-your-preferences"
+        >
+          update your preferences
+        </Link>
+        . Your details will be kept safe, check out our
+        <Link type="standard" href="https://www.comicrelief.com/privacy-policy">
+          privacy policy
+        </Link>
+        for more details.
+      </Text>
```
tomisaacmarshall, line 170: React (of JSX?) removes the spaces around these links I think

PR #311, opened 2020-06-24
Converting marketing preferences inputs to be 'uncontrolled' 

tomisaacmarshall:
In line with react-hook-form recommendations


gusliedke:
Nice, do we need a PR to test Schools Form with those changes?

tomisaacmarshall:
> Nice, do we need a PR to test Schools Form with those changes?

Yeah, I'll do that now

PR #318, opened 2020-07-07
feat: update modules

gusliedke:
BREAKING CHANGE: updagrading all modules

Type: patch

- [x] removing prettier
- [x] add same eslint config from CR . com
- [x] fix all lints
- [x] update all modules


gusliedke:
@tomisaacmarshall do you mind to have a look at
```TypeError: Cannot read property 'length' of undefined
(anonymous function)
./src/components/Atoms/TextInputWithDropdown/TextInputWithDropdown.js:104
  101 |   className={`TextInputWithDropdown ${className}`.trim()}
  102 |   onKeyDown={navigateOptions}
  103 | >
> 104 |   <Input
      | ^  105 |     {...inputProps}
  106 |     className="TextInputWithDropdown__input"
  107 |     ref={ref}
```
it's happening now when I upgraded styleguidist.

gusliedke:
A bit big but mainly eslint fixes and snapshot with different classes.

src/components/Atoms/Icons/Favourite.js
```diff
@@ -2,20 +2,20 @@ import React from 'react';
...
-    </svg>
-  );
-};
+const Favourite = ({
+  colour, theme, size, ...rest
+}) => (
+  <svg
+    {...rest}
+    width={size}
+    heiht={size}
```
AndyEPhipps, line 11: Not part of this PR admittedly, but there's a typo here.. not sure how much work it'd be to fix that

gusliedke: I can fix that 👍

PR #322, opened 2020-07-08
Bite-size Component

### LINKED ISSUE #321: Accordion Component
Related issue https://github.com/comicrelief/comicrelief-contentful/issues/656

image

```
two children prop

- First child (which is the Title in this case)
- Second child (which is the body in this case)

you click first child and it opens the body
```

LouiaR:
Fixes #321


gusliedke:
Hello, @Ruby-HB I'm not very sure about this Bite size component. We already have the new [Card](https://deploy-preview-322--cr-component-library.netlify.app/#cardds) which is exactly the same.

Maybe we could go through all pages in CR .com and start implementing the new components before creating those extra ones to see its usefulness.

Ruby-HB:
Hiya @gusliedke and @LouiaR I understand your points on this component. We will set some time aside soon to discuss the progression of components (I have made Chao aware of this too). 

We just want to be clear that there is a specific use case for this component. We will look into the design and make sure the functionality is distinct from the Card. Thanks, I'll get back to you with more info soon :)

gusliedke:
This will be done directly on CF only

PR #323, opened 2020-07-14
feat: update promo

gusliedke:
Type: patch

Fixes #264

### DIFF
--- a/src/components/Molecules/Promo/Promo.js
+++ b/src/components/Molecules/Promo/Promo.js
@@ -5,7 +5,6 @@ import Picture from '../../Atoms/Picture/Picture';
 import {
   Container,
   NudgeBottom,
-  NudgeTop,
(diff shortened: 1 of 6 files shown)


src/components/Molecules/Promo/Promo.js
```diff
@@ -5,53 +5,50 @@ import Picture from '../../Atoms/Picture/Picture';
 import {
   Container,
   NudgeBottom,
```
LouiaR, line 7: Did you try clip-path ?

gusliedke: No, clip-path is interesting .. looks like it has good support .. I will try it out!

LouiaR: Can you update font family of CTA as well

PR #327, opened 2020-08-05
Accordion Component

### LINKED ISSUE #321: Accordion Component

LouiaR:
Fixes #321 

https://deploy-preview-327--cr-component-library.netlify.app/#!/Accordion/1

### DIFF
--- a/src/components/Molecules/Accordion/Accordion.js
+++ b/src/components/Molecules/Accordion/Accordion.js
@@ -2,17 +2,27 @@ import React from 'react';
 import PropTypes from 'prop-types';
 import styled from 'styled-components';
 
+import spacing from '../../../theme/shared/spacing';
+import Text from '../../Atoms/Text/Text';
+
 const Container = styled.div`
   display: flex;
   position: relative;
   flex-direction: column;
   height: 100%;
   background: ${({ theme, backgroundColor }) => theme.color(backgroundColor)};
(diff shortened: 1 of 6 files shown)


gusliedke:
Title should be a node as well, so it can have different formats.

src/components/Molecules/Accordion/Accordion.js
```diff
@@ -0,0 +1,97 @@
...
+const Icon = styled.div`
+  display: flex;
+  justify-content: center;
+  align-content: center;
+`;
+
+const Copy = styled.div`
+  overflow: hidden;
+  visibility: none;
+  transition: all 0.2s linear;
```
AndyEPhipps, line 57: Silly, but  howsabout a nicer animation here? Looks a bit stiff using `linear`.. 'pull-up window blind' style?

`transition: all 0.2s cubic-bezier(0.21, 1.7, 0.83, 0.68) 0s;`

AndyEPhipps (approved):

src/components/Molecules/Accordion/Accordion.js
gusliedke: like it 👍

AndyEPhipps: 🎉

PR #334, opened 2020-08-11
feat: add 'xsmall' screen size to media helper

tomisaacmarshall:
This ok?

### DIFF
--- a/src/theme/shared/size.js
+++ b/src/theme/shared/size.js
@@ -1,4 +1,5 @@
 const screen = {
+  xsmall: '375px',
   small: '740px',
   medium: '1024px',
   large: '1440px'


gusliedke:
What's the new breakpoint for?

tomisaacmarshall:
> What's the new breakpoint for?

Just thought it'd be useful to be able to target v small screens?

gusliedke:
> > What's the new breakpoint for?
> 
> Just thought it'd be useful to be able to target v small screens?



> > What's the new breakpoint for?
> 
> Just thought it'd be useful to be able to target v small screens?

Only if we need to make the design to look different when on a small mobile screen vs large mobile screen. I don't think we want that. The idea is mobile first and the next breakpoint (design change) is just before tablet.

tomisaacmarshall:
> > > What's the new breakpoint for?
> > 
> > 
> > Just thought it'd be useful to be able to target v small screens?
> 
> > > What's the new breakpoint for?
> > 
> > 
> > Just thought it'd be useful to be able to target v small screens?
> 
> Only if we need to make the design to look different when on a small mobile screen vs large mobile screen. I don't think we want that. The idea is mobile first and the next breakpoint (design change) is just before tablet.

Ok!

PR #336, opened 2020-08-12
Add chevron icons

### LINKED ISSUE #335: Add chevron icon

LouiaR:
Fixes #335


gusliedke:
Can we use one `svg` and apply transform rotate for each direction?

LouiaR:
> Can we use one `svg` and apply transform rotate for each direction?

Do-able

PR #351, opened 2020-09-03
Review DS pages

gusliedke:
Type: patch

fixes: #352

### DIFF
--- a/src/components/Molecules/Promo/Promo.js
+++ b/src/components/Molecules/Promo/Promo.js
@@ -13,14 +13,15 @@ const Promo = ({
   imageSet,
   image,
   imageAltText,
-  children
+  children,
+  hasOverlay
 }) => {
   const hasImage = imageSet || false;
 
   return (
     <Container backgroundColor={backgroundColor}>
       {hasImage && (
-      <Media>
+      <Media hasOverlay={hasOverlay}>
         <Picture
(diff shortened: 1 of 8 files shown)


AndyEPhipps:
It doesn't look like something you've introduced _here_, but I've just noticed the weird font on the Promo link:

Screen Shot 2020-09-10 at 10 59 14

Do you think it's worth adding variants to the CL, so a hasOverlay and non-overlay version here? I figure it's a bit better in general to show _all_ possibilities (to help flag a breaking change).. although I guess it makes the CL a bit messy.. 🤔

gusliedke:
> It doesn't look like something you've introduced _here_, but I've just noticed the weird font on the Promo link:
> 
> Screen Shot 2020-09-10 at 10 59 14
> 
> Do you think it's worth adding variants to the CL, so a hasOverlay and non-overlay version here? I figure it's a bit better in general to show _all_ possibilities (to help flag a breaking change).. although I guess it makes the CL a bit messy.. 🤔

The btn text is because it's not wrapped on our Text component. I will add a Text in the demo btn.

The overlay we can switch on/off on the markup editor. We can click PROPS & METHODS to see all props available. I will add the option on test though.. I think it's valid to have a snap of no overlay 👍

AndyEPhipps (approved):

PR #372, opened 2020-10-19
feat: Update School Lookup component to track error count

tomisaacmarshall:
...and provide it to error handling callback so that parent components can handle errors more elegantly


tomisaacmarshall:
Unnecessary actually!

PR #374, opened 2020-10-19
Social event tracking

gusliedke:
Type: patch

Reviewed by Hiral. Track is working

### DIFF
--- a/src/components/Molecules/ShareButton/ShareButton.js
+++ b/src/components/Molecules/ShareButton/ShareButton.js
@@ -4,7 +4,7 @@ import Icon from '../../Atoms/SocialIcons/Icon/Icon';
 import popUpHelper from '../../../utils/ShareButton/sharePopUpHelper';
 import shareUrlHelper from '../../../utils/ShareButton/shareUrlHelper';
 import shareTracking from '../../../utils/ShareButton/shareTracking';
-import shareIcons from './assets/shareIcons';
+import ShareIcons from './assets/ShareIcons';
 
 import {
   Wrapper,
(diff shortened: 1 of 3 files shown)


src/utils/ShareButton/shareTracking.js
```diff
@@ -0,0 +1,10 @@
+const shareTracking = channel => {
+  const dataLayer = window.dataLayer || [];
```
tomisaacmarshall, line 2: This needs to be a reference to the global `dataLayer` object

Something like this would work

```
if (!window.dataLayer) {
  window.dataLayour = [];
}

window.dataLayer.push({
    event: 'btn_social-share',
    action: channel
});
```

At present if `window.dataLayer` isn't already defined, we'd just be adding the event to a new array which will be chucked away at the end of the function.

src/components/Molecules/ShareButton/ShareButton.js
```diff
@@ -15,6 +16,8 @@ import {
 const handleShare = (e, typeOfShare, urlToShare) => {
   e.preventDefault();
 
+  ShareTracking(typeOfShare.toLowerCase());
```
tomisaacmarshall, line 19: I'd probably do the lowercasing in the `ShareTracking` function. That way we don't have to just _know_ that this function wants a lowercased value.

src/utils/ShareButton/shareTracking.js
```diff
@@ -0,0 +1,10 @@
...
+
+  dataLayer.push({
+    event: 'btn_social-share',
+    action: channel
```
tomisaacmarshall: Would it be useful to record the URL as well?

src/components/Molecules/ShareButton/ShareButton.js
tomisaacmarshall: Sorry another minor point - would it be more conventional to lowercase the first letter in `ShareTracking` and `ShareUrlHelper`? Def not an expert on JS naming conventions but thought capital first letter implied a class (or React component) rather than a function?

gusliedke: Left as the other were in capital but I will change them to correct convention 👍

src/utils/ShareButton/shareTracking.js
gusliedke: @phiral how about we record `url` as well?

phiral: @gusliedke no need for the URL, I can grab that from the history change event.

phiral: When I send the info in Google Analytics I send the page URL with the event. So no need to specify the url.

tomisaacmarshall (approved):

PR #375, opened 2020-10-20
Fix/get tests to pass

tomisaacmarshall:


KrupaPammi (approved):

cypress/integration/components/Molecules/SchoolLookup.spec.js
```diff
@@ -1,107 +1,124 @@
...
+        .contains('View Code')
+        .as('viewCodeBtn');
+    });
 
-        it('should verify schools lookup functionality', () => {
-            //verify error message for non school postcode
+    it('should verify schools lookup functionality', () => {
+      // Check if the API is up and running before running the tests.
+      cy.request({
+        url: 'https://lookups.sls.comicrelief.com/schools/lookup?query=NW5',
```
KrupaPammi, line 81: Is there any reason we are using production lookup url here?

tomisaacmarshall: Yeah. Need to use the same URL as the component. Which just uses the hardcoded prod URL. Not sure if it's possible to vary it depending on environment in the component library - depends if/how the components are bundled I guess.

Don't think there's any reason to use the staging url for this particular API as it's not creating anything

PR #390, opened 2020-11-18
fix: nav menu mobile

gusliedke:
Type: patch

Fixes #

- Briefly describe the changes in this PR
- Do you have some test asserting this behavior?

### DIFF
--- a/src/components/Organisms/Header/Nav/Nav.js
+++ b/src/components/Organisms/Header/Nav/Nav.js
@@ -36,11 +36,11 @@ const MainNav = ({ navItems }) => {
   const toggleSubMenu = (item, group) => event => {
     // Check if navLink element has more than one subNav item
     const checkSubnav = group && group.length > 1;
-    if (checkSubnav) {
-      console.log(group);
-      event.preventDefault();
-      setIsSubMenuOpen({ [item]: !isSubMenuOpen[item] });
-    }
+    console.log(group);
+    event.preventDefault();
+    setIsSubMenuOpen({ [item]: !isSubMenuOpen[item] });
+    // if (checkSubnav) {
+    // }
   };
 
   // Handle tab key on menu nav
--- a/src/components/Organisms/Header/Nav/Nav.js
+++ b/src/components/Organisms/Header/Nav/Nav.js
@@ -110,19 +110,17 @@ const MainNav = ({ navItems }) => {
                     </NavLink>
                   </Text>
                 ) : (
-                  <Text>
-                    <NavLink
-                      href={thisUrl}
-                      inline
-                      aria-expanded={!!isSubMenuOpen[group.id]}
-                      aria-haspopup={hasPopUp}
-                      onClick={toggleSubMenu(group.id, group.links)}
-                      onKeyUp={keyPressed(group.title)}
-                      role="button"
-                    >
-                      {thisFirstChild.title}
-                    </NavLink>
-                  </Text>
+                  <NavLink
+                    href={thisUrl}
+                    inline
+                    aria-expanded={!!isSubMenuOpen[group.id]}
+                    aria-haspopup={hasPopUp}
+                    onClick={toggleSubMenu(group.id, group.links)}
+                    onKeyUp={keyPressed(group.title)}
+                    role="button"
+                  >
+                    {thisFirstChild.title}
+                  </NavLink>
                 )}
                 {/* Second level of the navigation (ul tag): Child(ren) */}
                 {hasSubMenu && (
--- a/src/components/Organisms/Header/Nav/Nav.js
+++ b/src/components/Organisms/Header/Nav/Nav.js
@@ -36,7 +36,7 @@ const MainNav = ({ navItems }) => {
   const toggleSubMenu = (item, group) => event => {
     // Check if navLink element has more than one subNav item
     const checkSubnav = group && group.length > 1;
-    if ((isMobile || isTouch) && checkSubnav) {
+    if (checkSubnav) {
       event.preventDefault();
       setIsSubMenuOpen({ [item]: !isSubMenuOpen[item] });
     }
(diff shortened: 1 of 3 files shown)


src/components/Organisms/Header/Nav/Nav.js
```diff
@@ -110,19 +102,17 @@ const MainNav = ({ navItems }) => {
...
+                    href={hasPopUp ? '#' : thisUrl}
+                    inline
+                    aria-expanded={!!isSubMenuOpen[group.id]}
+                    aria-haspopup={hasPopUp}
+                    onClick={hasPopUp ? e => toggleSubMenu(e, group.id) : null}
+                    onKeyUp={keyPressed(group.title)}
+                    role="button"
+                  >
+                    {thisFirstChild.title}
+                  </NavLink>
```
tomisaacmarshall, line 115: I know it's not due to your changes but I find this whole block very hard to parse:

```
{!isMobile ? (
                  <Text>
                    <NavLink
                      href={thisUrl}
                      inline
                      rel={relNoopener}
                      aria-haspopup={hasPopUp}
                      onKeyUp={keyPressed(group.title)}
                    >
                      {thisFirstChild.title}
                    </NavLink>
                  </Text>
                ) : (
                  <NavLink
                    href={hasPopUp ? '#' : thisUrl}
                    inline
                    aria-expanded={!!isSubMenuOpen[group.id]}
                    aria-haspopup={hasPopUp}
                    onClick={hasPopUp ? e => toggleSubMenu(e, group.id) : null}
                    onKeyUp={keyPressed(group.title)}
                    role="button"
                  >
                    {thisFirstChild.title}
                  </NavLink>
                )}
```

Do we need the ternary or could just have one `NavLink` element with conditional props?

gusliedke: I've taken some unnecessary logic already .. I got a bit apprehensive touching it more .. that's definitely something we need to get back to. I'm working with Ruby on the new Header, hopefully we can replace it soon.

src/components/Organisms/Header/Nav/Nav.js
```diff
@@ -57,14 +53,10 @@ const MainNav = ({ navItems }) => {
   };
 
   useEffect(() => {
-    const isTouchDevice = (typeof window !== 'undefined' && 'ontouchstart' in window)
-      || window.navigator.msMaxTouchPoints > 0;
-
     const width = typeof window !== 'undefined' ? window.innerWidth : null;
```
tomisaacmarshall, line 56: Not anything you've added but could remove this check while we're here as `useEffect` only ever runs in the browser

src/components/Organisms/Header/Nav/Nav.js
tomisaacmarshall: Ok well at the very least ... would be good to flip the ternary if poss? 

I always find `!something ? ... : ...` hard to follow

src/components/Organisms/Header/Nav/Nav.js
```diff
@@ -57,14 +53,10 @@ const MainNav = ({ navItems }) => {
...
 
   useEffect(() => {
-    const isTouchDevice = (typeof window !== 'undefined' && 'ontouchstart' in window)
-      || window.navigator.msMaxTouchPoints > 0;
-
-    const width = typeof window !== 'undefined' ? window.innerWidth : null;
+    const width = window.innerWidth;
     // Detect window screen size
     setIsMobile(width < sizes.nav);
     // Detect if it is a touch device
```
tomisaacmarshall, line 59: This comment needs removing

src/components/Organisms/Header/Nav/Nav.js
```diff
@@ -97,28 +89,26 @@ const MainNav = ({ navItems }) => {
...
-                      href={thisUrl}
-                      inline
-                      rel={relNoopener}
-                      aria-haspopup={hasPopUp}
-                      onKeyUp={keyPressed(group.title)}
-                    >
-                      {thisFirstChild.title}
-                    </NavLink>
-                  </Text>
+                {isMobile ? (
```
tomisaacmarshall, line 92: As I say, I find it hard to parse, but I _think_ larger touch devices with screens wider than `sizes.nav` e.g. tablets won't see the dropdowns

However, guessing the touch detection was removed because it didn't actually work? If so maybe this is something to look at later?

src/components/Organisms/Header/Nav/Nav.js
tomisaacmarshall: Thanks!

src/components/Organisms/Header/Nav/Nav.js
gusliedke: For screens larger than `nav` desktop header will show.. don't think we need to worry now for touch screens bigger than 1150px.

tomisaacmarshall: Was thinking of ipads in landscape but guess it's not gonna be many people

gusliedke: yeah don't think so, still.. the open/close is only for the mobile menu (burger menu) .. at this size 1150 desktop header will display.

tomisaacmarshall: Yeh I totally get that. I just meant that e.g an ipad pro in landscape would show the desktop header despite being a touch device. So users wouldn't be able to get to the dropdowns, as it has no hover state. Ideally whether to employ touch behaviour would be primarily based on whether it's a touch device rather than viewport width. But appreciate it's an edge case and detecting touch would make it more complicated!

tomisaacmarshall (approved):
Looks good. FYI one other small thing I've noticed about the header which I'll leave at your discretion ... the `cursor: pointer` is attached to the containing `li` element which is larger than the `a` element, so non-clickable space looks clickable.

src/components/Organisms/Header/Nav/Nav.js
gusliedke: ah ok ok. yeah, I believe you still can open the dropdown on touch (hard touch) .. but same if I open CR on surface or any laptop with touch screen it will show me the desktop version. Something to be aware in the future!

PR #394, opened 2020-11-23
feat: Add Label component

tomisaacmarshall:
To reduce repeated code across form controls and ensure they have consistent label styling

### DIFF
--- a/src/components/Atoms/FieldWrapper/FieldWrapper.js
+++ b/src/components/Atoms/FieldWrapper/FieldWrapper.js
@@ -1,0 +1,69 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import styled from 'styled-components';
+
+import Text from '../Text/Text';
+import ErrorText from '../ErrorText/ErrorText';
+import spacing from '../../../theme/shared/spacing';
+import hideVisually from '../../../theme/shared/hideVisually';
+
+const Label = styled.label`
+  display: flex;
+  flex-direction: column;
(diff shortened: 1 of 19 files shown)


gusliedke:
Maybe call it **Label** instead?

tomisaacmarshall:
> Maybe call it **Label** instead?

I nearly did but as it also displays the error message under the field, thought it wasn't quite right

tomisaacmarshall:
> > Maybe call it **Label** instead?
> 
> I nearly did but as it also displays the error message under the field, thought it wasn't quite right

Perhaps I should change it so the error message goes back into the input/select/textarea component, and this is just the label?

gusliedke:
> > > Maybe call it **Label** instead?
> > 
> > 
> > I nearly did but as it also displays the error message under the field, thought it wasn't quite right
> 
> Perhaps I should change it so the error message goes back into the input/select/textarea component, and this is just the label?

Yeah I think just Label would be better .. and Error msg stay flexible where it displays

tomisaacmarshall:
> > > > Maybe call it **Label** instead?
> > > 
> > > 
> > > I nearly did but as it also displays the error message under the field, thought it wasn't quite right
> > 
> > 
> > Perhaps I should change it so the error message goes back into the input/select/textarea component, and this is just the label?
> 
> Yeah I think just Label would be better .. and Error msg stay flexible where it displays

@gusliedke agreed. Will update

gusliedke (approved):

PR #396, opened 2020-11-23
Remove SchoolLookup tests that keep failing

tomisaacmarshall:
I think we just need to get rid of these tests

They are costing too much time when they keep failing

### DIFF
--- a/cypress/integration/components/Molecules/SchoolLookup.spec.js
+++ b/cypress/integration/components/Molecules/SchoolLookup.spec.js
@@ -1,6 +1,6 @@
 describe('School Lookup component', () => {
   before(() => {
-        // go to school lookup component
+    // go to school lookup component
     cy.visit('/#schoollookup');
   });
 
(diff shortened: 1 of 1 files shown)


cypress/integration/components/Molecules/SchoolLookup.spec.js
```diff
@@ -74,51 +74,5 @@ describe('School Lookup component', () => {
...
-            cy.get('@container')
-              .find('#school-lookup')
-              .type('se1 7tp')
-              .get('@container')
-              .find('[data-test="error-message"]')
-              .contains('Sorry, there was an unexpected error. Please try again');
-          } else {
-            // verify error message for non school postcode
-            cy.get('@container')
-              .find('#school-lookup')
```
KrupaPammi, line 95: Looks like the lookup test is not right here, there are no manual fields that are displayed for `else` block to execute when lookup status >= 400 

Screen Shot 2020-11-23 at 16 49 14

tomisaacmarshall: Manual fields? Sorry I don't follow

KrupaPammi: What I mean is, when lookup status is not 200, the field itself doesn't showup for rest of the code to execute hence it fails, removing the else blocks works

KrupaPammi (approved):

PR #398, opened 2020-11-25
feat: Accordion updates

tomisaacmarshall:
Updated the Accordion component as I needed to be able to style it with a different background colour, (so added `rest` props which enables me to use it with `styled()`)

Also removed some unused code & tweaked the CSS slightly to make the whole bar clickable (i.e. including the padding)


gusliedke (approved):
👍

PR #412, opened 2020-12-02
Countdown

gusliedke:
### Countdown

image

https://deploy-preview-412--cr-component-library.netlify.app/#!/Countdown

### DIFF
--- a/src/components/Molecules/Countdown/Countdown.js
+++ b/src/components/Molecules/Countdown/Countdown.js
@@ -15,8 +15,8 @@ const Countdown = ({
     seconds: '00'
   });
 
-  const countDownDate = moment(endDate);
-  countDownDate.diff(moment(), 'seconds');
+  let countDownDate = moment(endDate);
+  // countDownDate.diff(moment(), 'seconds');
 
   useEffect(() => {
     const interval = setInterval(() => {
@@ -29,13 +29,13 @@ const Countdown = ({
       const seconds = timeRemaining.format('ss');
 
       setCountdownTime({
-        days,
-        hours,
-        minutes,
-        seconds
+        days: days - 1,
+        hours: hours - 1,
+        minutes: minutes,
+        seconds: seconds
       });
 
-      if (timeRemaining < 0) {
+      if (countDownDate.diff(now, 'seconds') < 1) {
         setCountdownHasEnded(true);
       }
     }, 1000);
--- a/src/components/Molecules/Countdown/Countdown.js
+++ b/src/components/Molecules/Countdown/Countdown.js
@@ -3,34 +3,33 @@ import PropTypes from 'prop-types';
 import Text from '../../Atoms/Text/Text';
 import { Wrapper, Digits } from './Countdown.style';
 
-
-const Countdown = ({endDate, color, endMessage, introMessage}) => {
+const Countdown = ({
+  endDate, color, endMessage, introMessage
+}) => {
   const [countdownEnd, setCountdownEnd] = useState(false);
   const [countdownTime, setCountdownTime] = useState({});
-  
+
   const countDownDate = new Date(endDate).getTime();
   // Keep number with two digits
-  const formatNumber = number => {
-    return ("0" + number).slice(-2);
-  }
+  const formatNumber = number => (`0${number}`).slice(-2);
 
   useEffect(() => {
     const interval = setInterval(() => {
       const now = new Date().getTime();
       const distance = countDownDate - now;
-  
+
       const days = formatNumber(Math.floor(distance / (1000 * 60 * 60 * 24)));
       const hours = formatNumber(Math.floor((distance % (1000 * 60 * 60 * 24)) / (1000 * 60 * 60)));
       const minutes = formatNumber(Math.floor((distance % (1000 * 60 * 60)) / (1000 * 60)));
       const seconds = formatNumber(Math.floor((distance % (1000 * 60)) / 1000));
-  
+
       setCountdownTime({
-        days: days,
-        hours: hours,
-        minutes: minutes,
-        seconds: seconds
+        days,
+        hours,
+        minutes,
+        seconds
       });
-  
+
       if (distance < 0) {
         setCountdownEnd(true);
       }
@@ -83,7 +82,7 @@ const Countdown = ({endDate, color, endMessage, introMessage}) => {
       </Wrapper>
     </>
   );
-  
+
   if (endMessage == null) {
     return !countdownEnd ? CountDownClock : null;
   }
@@ -96,12 +95,12 @@ Countdown.protoTypes = {
   color: PropTypes.string,
   endMessage: PropTypes.node,
   introMessage: PropTypes.node
-}
+};
 
 Countdown.defaultProps = {
-  color: "black",
+  color: 'black',
   endMessage: null,
   introMessage: null
-}
+};
 
 export default Countdown;
--- a/src/components/Molecules/Countdown/Countdown.js
+++ b/src/components/Molecules/Countdown/Countdown.js
@@ -1,0 +1,61 @@
+import React, { useState, useEffect } from 'react';
+import PropTypes from 'prop-types';
+import Text from '../../Atoms/Text/Text';
+
+
+const Countdown = ({endDate, color, endMessage}) => {
+  const [countdownEnd, setCountdownEnd] = useState(false);
+  const [countdownTime, setCountdownTime] = useState("");
+  
+  const countDownDate = new Date(endDate).getTime();
+  // Keep number with two digits
+  const formatNumber = number => {
+    return ("0" + number).slice(-2);
+  }
+
+  useEffect(() => {
+    const interval = setInterval(() => {
+      const now = new Date().getTime();
+      const distance = countDownDate - now;
+  
+      const days = formatNumber(Math.floor(distance / (1000 * 60 * 60 * 24)));
+      const hours = formatNumber(Math.floor((distance % (1000 * 60 * 60 * 24)) / (1000 * 60 * 60)));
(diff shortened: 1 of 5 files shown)


gusliedke:
Leaving CY tests to be create on Challenge page.

src/components/Molecules/Countdown/Countdown.js
```diff
@@ -0,0 +1,106 @@
...
+      if (distance < 0) {
+        setCountdownEnd(true);
+      }
+    }, 1000);
+    return () => clearInterval(interval);
+  }, [countDownDate]);
+
+  const CountDownClock = (
+    <>
+      {introMessage && introMessage}
```
tomisaacmarshall, line 42: This is the same as just `{introMessage}` isn't it?

src/components/Molecules/Countdown/Countdown.js
```diff
@@ -0,0 +1,106 @@
...
+  const [countdownTime, setCountdownTime] = useState({});
+
+  const countDownDate = new Date(endDate).getTime();
+  // Keep number with two digits
+  const formatNumber = number => (`0${number}`).slice(-2);
+
+  useEffect(() => {
+    const interval = setInterval(() => {
+      const now = new Date().getTime();
+      const distance = countDownDate - now;
```
tomisaacmarshall, line 19: How about `timeRemaining`?

src/components/Molecules/Countdown/Countdown.js
```diff
@@ -0,0 +1,106 @@
+import React, { useState, useEffect } from 'react';
+import PropTypes from 'prop-types';
+import Text from '../../Atoms/Text/Text';
+import { Wrapper, Digits } from './Countdown.style';
+
+const Countdown = ({
+  endDate, color, endMessage, introMessage
+}) => {
+  const [countdownEnd, setCountdownEnd] = useState(false);
```
tomisaacmarshall, line 9: Small point but I reckon `countdownHasEnded` would make things a bit easier for follow

src/components/Molecules/Countdown/Countdown.js
```diff
@@ -0,0 +1,106 @@
...
+        </Digits>
+      </Wrapper>
+    </>
+  );
+
+  if (endMessage == null) {
+    return !countdownEnd ? CountDownClock : null;
+  }
+
+  return countdownEnd ? endMessage : CountDownClock;
```
tomisaacmarshall, line 90: Finding this section a bit confusing.

I think the first return is actually redundant and can just be removed as in effect it's identical to the second one. 

All it seems to be doing differently is returning `null` instead of `endMessage` when `countdownEnd` is truthy - but only when `endMessage` is already equivalent to null

src/components/Molecules/Countdown/Countdown.js
```diff
@@ -0,0 +1,106 @@
...
+      });
+
+      if (distance < 0) {
+        setCountdownEnd(true);
+      }
+    }, 1000);
+    return () => clearInterval(interval);
+  }, [countDownDate]);
+
+  const CountDownClock = (
```
tomisaacmarshall, line 40: The loading state doesn't look great ... would it be better to start with '00 : 00 : 00 : 00', or just not render the clock until `useEffect` has executed and set the time?

src/components/Molecules/Countdown/Countdown.js
```diff
@@ -0,0 +1,106 @@
...
+  }
+
+  return countdownEnd ? endMessage : CountDownClock;
+};
+
+Countdown.protoTypes = {
+  endDate: PropTypes.string.isRequired,
+  color: PropTypes.string,
+  endMessage: PropTypes.node,
+  introMessage: PropTypes.node
```
tomisaacmarshall, line 97: Is there any point in having this prop, given that we don't apply any styling to it?
Couldn't we just use `<Text>` before the component?

src/components/Molecules/Countdown/Countdown.js
```diff
@@ -0,0 +1,106 @@
...
+
+  useEffect(() => {
+    const interval = setInterval(() => {
+      const now = new Date().getTime();
+      const distance = countDownDate - now;
+
+      const days = formatNumber(Math.floor(distance / (1000 * 60 * 60 * 24)));
+      const hours = formatNumber(Math.floor((distance % (1000 * 60 * 60 * 24)) / (1000 * 60 * 60)));
+      const minutes = formatNumber(Math.floor((distance % (1000 * 60 * 60)) / (1000 * 60)));
+      const seconds = formatNumber(Math.floor((distance % (1000 * 60)) / 1000));
```
tomisaacmarshall, line 24: Is it worth unit testing this stuff or using a library e.g. [date-fns](https://date-fns.org/docs/)?

src/components/Molecules/Countdown/Countdown.js
```diff
@@ -0,0 +1,106 @@
...
+
+      setCountdownTime({
+        days,
+        hours,
+        minutes,
+        seconds
+      });
+
+      if (distance < 0) {
+        setCountdownEnd(true);
```
tomisaacmarshall, line 34: Just checking, we definitely want to hide the countdown when it's finished, rather than displaying `00:00:00:00` (and letting the parent component decide if it wants to hide it) ?

(Not sure where / how this component will be used)

tomisaacmarshall (changes requested):

src/components/Molecules/Countdown/Countdown.js
gusliedke: The problem here is that we want Intro text to go away when countdown is finished. If we don't have it as part of this component we need remove the copy in the CMS manually. Ideally we would have `node` but for this version we can simply have a text.

src/components/Molecules/Countdown/Countdown.js
gusliedke: Cool done '00 : 00 : 00 : 00' as initial state. 👍

src/components/Molecules/Countdown/Countdown.js
tomisaacmarshall: Ahh of course. Makes sense!

src/components/Molecules/Countdown/Countdown.js
gusliedke: I thought about using `momentjs` (we use it in cr.com) but I didn't want to add another dependency.. Do you think is a risk leaving as it is?

src/components/Molecules/Countdown/Countdown.js
gusliedke: yeah right, I think I got it better now.

gusliedke:
weird why I'm getting
```  7:3   error  'endDate' is missing in props validation       react/prop-types
  7:12  error  'color' is missing in props validation         react/prop-types
  7:19  error  'endMessage' is missing in props validation    react/prop-types
  7:31  error  'introMessage' is missing in props validation  react/prop-types```

File: src/components/Molecules/Countdown/Countdown.js (hunk 7 above)
Comment (tomisaacmarshall) [reply src/components/Molecules/Countdown/Countdown.js]: I think it should be tested tbh. So that we can be confident it works, and confirm that it works as expected across time zones and with daylight saving changes.

You could extract the calculation part into a function outside of the component and export that to test it.

& I think using a library could make it a lot easier to follow and maintain. I'm not sure whether another dependency is much of an issue? `date-fns` and `moment` are both included in `yarn.lock` so they are already dependencies of depedencies

Comment (tomisaacmarshall) [reply src/components/Molecules/Countdown/Countdown.js]: TBF it looks like they are dependencies of cypress, so I guess they aren't being used by the components. But to my mind, the benefits of using them outweigh any costs, although I could definitely be wrong...

Comment (gusliedke) [reply src/components/Molecules/Countdown/Countdown.js]: I will try and use `moment`.

File: src/components/Molecules/Countdown/Countdown.js (hunk 4 above)
Comment (tomisaacmarshall) [reply src/components/Molecules/Countdown/Countdown.js]: It's easier to see if you flip the first ternary

```

  if (endMessage == null) {
    return countdownEnd ? null : CountDownClock;
  }

  return countdownEnd ? endMessage : CountDownClock;
```

The second return would return `null` for `endMessage` anyway if the condition `endMessage === null` is met

File: src/components/Molecules/Countdown/Countdown.js (hunk 7 above)
Comment (tomisaacmarshall) [reply src/components/Molecules/Countdown/Countdown.js]: Cool. The reason I mentioned date-fns btw is that it's more lightweight, you can just import the individual functions you need, so I _think_ it would lead to a smaller package

**Comment by gusliedke** [comment]
> weird why I'm getting
> 
> ```
>   7:12  error  'color' is missing in props validation         react/prop-types
>   7:19  error  'endMessage' is missing in props validation    react/prop-types
>   7:31  error  'introMessage' is missing in props validation  react/prop-types
> ```

@tomisaacmarshall any idea why this start failing now? Props validation are there.

File: src/components/Molecules/Countdown/Countdown.js [hunk 9]
```diff
@@ -0,0 +1,106 @@
...
+          <Text color={color} size="xs" uppercase>
+            seconds
+          </Text>
+        </Digits>
+      </Wrapper>
+    </>
+  );
+};
+
+Countdown.protoTypes = {
```
tomisaacmarshall, line 93: Typo here `Countdown.protoTypes`

gusliedke: 😅 👍

src/components/Molecules/Countdown/Countdown.js
gusliedke: Countdown will hide when it's finished. If there is a endmessage it will display it if not it will be null.

PR #417, opened 2021-02-16
feat: adds AtSign icon component

tomisaacmarshall:


gusliedke:
Do you mind to add an example on `https://github.com/comicrelief/component-library/blob/master/src/components/Atoms/Link/Link.md#iconography`

gusliedke (approved):

PR #422, opened 2021-04-08
upgrade modules

gusliedke:
Type: patch

Fixes #

- Briefly describe the changes in this PR
- Do you have some test asserting this behavior?


AndyEPhipps:
@gusliedke, I've got this MP fix working within my local Contentful, just having to manually pass the formContext as prop

There's likely an underlying issue as to why `useFormContext` isn't working with components pulled-in from the CL, but in the interests of time/sanity (and the fact that Prize stuff is finishing shortly anyways), this seems like a reasonable solution to me.

I'll implement the associated update in your Contentful PR now

gusliedke:
> @gusliedke, I've got this MP fix working within my local Contentful, just having to _manually_ pass the formContext as a prop here now.
> 
> There's likely an underlying issue as to why `useFormContext` isn't working with components pulled-in from the CL, but in the interests of time/sanity (and the fact that Prize stuff is finishing shortly anyways), this seems like a reasonable solution to me.
> 
> I'll implement the associated update in your Contentful PR now

Thanks, Andy. Our last Prize just closed. We have time now until new Prizes to fix it or find a better solution. 👍

src/components/Molecules/CardDs/CardDs.js
```diff
@@ -139,7 +139,7 @@ const CardDs = ({
     return null;
   };
 
-  const external = target === 'blank' ? 'noopener noreferrer' : null;
+  const external = target === 'blank' ? 'noopener' : null;
```
AndyEPhipps, line 142: I take it `noreferrer` has been removed for SEO reasons?

src/components/Molecules/SingleMessageDS/SingleMessageDs.js
```diff
@@ -63,7 +63,7 @@ const SingleMessageDs = ({
 
   const icon = linkIcon || (target === 'blank' ? <External /> : <Internal />);
 
-  const external = target === 'blank' ? 'noopener noreferrer' : null;
+  const external = target === 'blank' ? 'noopener' : null;
```
AndyEPhipps, line 66: Ditto here?

src/components/Organisms/EmailSignUp/EmailSignUp.js
```diff
@@ -75,7 +75,7 @@ const EmailSignUp = ({
             value={value}
             onChange={event => setValue(event.target.value)}
           />
-          <ButtonWrapper backgroundColor={backgroundColor}>
+          <ButtonWrapper>
```
AndyEPhipps, line 78: Why don't we need this any more?

As an aside, we should probably be making these kind of adjustments in a separate PR, as it kinda muddies the water for a PR that's supposed to just cover module upgrades.

AndyEPhipps (approved):

src/components/Molecules/CardDs/CardDs.js
gusliedke: yeah you can you either not both.

src/components/Molecules/SingleMessageDS/SingleMessageDs.js
gusliedke: ditto

src/components/Organisms/EmailSignUp/EmailSignUp.js
gusliedke: No need of background colour it is wrong just a clean up.

PR #439, opened 2021-08-25
feat: upgrade no breaking modules

gusliedke:
Updating a few module

### DIFF
--- a/package.json
+++ b/package.json
@@ -21,10 +21,10 @@
     "url": "https://github.com/comicrelief/component-library.git"
   },
   "dependencies": {
-    "@babel/cli": "^7.13.14",
+    "@babel/cli": "^7.14.8",
     "@hookform/resolvers": "^1.3.4",
     "axios": "^0.21.1",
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -81,7 +81,7 @@
     "eslint-plugin-react": "^7.24.0",
     "eslint-plugin-react-hooks": "^4.0.5",
     "husky": "^7.0.2",
-    "jest": "^27.0.6",
+    "jest": "^26.1.0",
```
AndyEPhipps, line 84: What about Jest was breaking? Isn't it worth trying to address now because we'll have to update it eventually anyways?

gusliedke: It was an issue with `react-test-renderer` .. it got a bit messy with all tests and snaps updates I had to do and still couldn't get all errors sorted. I'd prefer to tackle it in a separated PR and leave this just with the Dependabot updates.

AndyEPhipps: Understood!

AndyEPhipps (approved):

PR #445, opened 2021-11-16
feat: new prop to override footer link target attr values

### LINKED ISSUE #444: Add new 'target' override prop to Footer nav links
Within the new Prefs centre, I'm pulling in the Component Library footer (as used on CR.com), and the Footer links in via the https://content.sls.comicrelief.com/footer end point. 

The only problem is, the whitelist functionality sees this as an internal link, and opens it in the same tab.

As such, I'm going to extend the footer to add a optional prop to force a `target="_blank"` attribute on all footer links, overriding the whitelist.

AndyEPhipps:
Type: feature

Fixes #444 

Basically just adds a new optional prop in to the footer menu, that can be picked up by the Link component, giving us the option to override the target value. Associated Issue copy here for your convenience:

> Within the new Prefs centre, I'm pulling in the Component Library footer (as used on CR.com), and the Footer links in via the https://content.sls.comicrelief.com/footer end point.
> 
> The only problem is, the whitelist functionality sees this as an internal link, and opens it in the same tab.
> 
> As such, I'm going to extend the footer to add a optional prop to force a target="_blank" attribute on all footer links, overriding the whitelist.

### DIFF
--- a/src/components/Atoms/Link/Link.js
+++ b/src/components/Atoms/Link/Link.js
@@ -3,7 +3,7 @@ import PropTypes from 'prop-types';
 
 import StyledLink, { HelperText, IconWrapper } from './Link.style';
 import whiteListed from '../../../utils/whiteListed';
-import { domainRegEx, getDomain } from '../../../utils/internalLinkHelper';
+import { getDomain } from '../../../utils/internalLinkHelper';
 
 let window = '';
 
@@ -33,11 +33,10 @@ const Link = ({
      * OR has the same domain as the current page, don't open
      * in a new tab
      */
-    const isNewTab = linkDomain !== '' && (currentDomain !== linkDomain);
+    const isExternalLink = linkDomain !== '' && (currentDomain !== linkDomain);
 
-    const isExternalLink = domainRegEx.test(href);
     const isWhiteListed = whiteListed(href);
-    window = isNewTab && (isExternalLink || !isWhiteListed) ? '_blank' : '_self';
+    window = isExternalLink && !isWhiteListed ? '_blank' : '_self';
   } else {
     window = target === 'blank' ? '_blank' : '_self';
   }
--- a/src/components/Atoms/Link/Link.js
+++ b/src/components/Atoms/Link/Link.js
@@ -3,10 +3,7 @@ import PropTypes from 'prop-types';
 
 import StyledLink, { HelperText, IconWrapper } from './Link.style';
 import whiteListed from '../../../utils/whiteListed';
-
-const domainRegEx = new RegExp(
-  '(?:[a-z0-9](?:[a-z0-9-]{0,61}[a-z0-9])?\\.)+[a-z0-9][a-z0-9-]{0,61}[a-z0-9]'
-);
+import { domainRegEx, getDomain } from '../../../utils/internalLinkHelper';
 
 let window = '';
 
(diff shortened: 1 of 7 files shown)


gusliedke:
"The only problem is, the whitelist functionality sees this as an internal link, and opens it in the same tab."

I'm wondering, could we verify if the link "lives" in the same url as the `internal link`? If so, opens in the same tab if different changes target to `_blank`.

ex. footer link `cr.com/terms` but link is in `markpref.cr.com` (origin different from link destination) change target to `_blank`.

AndyEPhipps:
> "The only problem is, the whitelist functionality sees this as an internal link, and opens it in the same tab."
> 
> I'm wondering, could we verify if the link "lives" in the same url as the `internal link`? If so, opens in the same tab if different changes target to `_blank`.
> 
> ex. footer link `cr.com/terms` but link is in `markpref.cr.com` (origin different from link destination) change target to `_blank`.

So comparing the Link href to the _current_ domain the Footer is used in? 

Hm yeah, not a bad idea; will give it a go

src/components/Atoms/Link/Link.js
```diff
@@ -3,10 +3,7 @@ import PropTypes from 'prop-types';
 
 import StyledLink, { HelperText, IconWrapper } from './Link.style';
 import whiteListed from '../../../utils/whiteListed';
-
-const domainRegEx = new RegExp(
-  '(?:[a-z0-9](?:[a-z0-9-]{0,61}[a-z0-9])?\\.)+[a-z0-9][a-z0-9-]{0,61}[a-z0-9]'
-);
+import { domainRegEx, getDomain } from '../../../utils/internalLinkHelper';
```
AndyEPhipps, line 6: Figured I'd export this to reduce the duplication

src/components/Atoms/Link/Link.js
```diff
@@ -27,9 +24,19 @@ const Link = ({
    * this is an internal link OR on our whitelist before making it a '_self' link
    */
   if (target === null) {
-    const isExternalLink = domainRegEx.test(href);
```
AndyEPhipps, line 30: I've removed this here as the same functionality is replaced (and improved) by the new code

src/components/Atoms/Link/Link.js
```diff
@@ -27,9 +24,19 @@ const Link = ({
...
+    // Use our helper function to determine the raw domains to compare
+    const currentDomain = getDomain(document.location.host);
+    const linkDomain = getDomain(href);
+
+    /**
+     * If the link has no domain supplied (likely '/' or '#')
+     * OR has the same domain as the current page, don't open
+     * in a new tab
+     */
+    const isExternalLink = linkDomain !== '' && (currentDomain !== linkDomain);
```
AndyEPhipps, line 36: Note, I'm keeping the same `isExternalLink` var name but it _has_ changed.

src/components/Atoms/Link/Link.js
```diff
@@ -46,7 +53,7 @@ const Link = ({
       underline={underline}
     >
       {children}
-      {target === 'blank' && <HelperText>(opens in new window)</HelperText>}
+      {window === '_blank' && <HelperText>(opens in new window)</HelperText>}
```
AndyEPhipps, line 56: Might as well tie this into our logic; `target` has the opportunity to be `null` in the prior code

AndyEPhipps:
Had to do a few slightly fiddly things to capture all usecases, but this seems pretty solid to me:

https://deploy-preview-445--cr-component-library.netlify.app/#footer

The whitelisting still works fine (where a link on a predefined domain will open in the same tab), as with the "Test whitelisted external link"

Opens in a new tab for "Test non-whitelisted external link"

Opens in the same tab for a proper internal link "Contact us" (you can see the path is just 'contact-us' in the [code](https://github.com/comicrelief/component-library/blob/master/src/components/Organisms/Footer/data/data.js#L30) )

gusliedke (approved):

PR #462, opened 2022-05-03
fix: undesirable image styling when no link is present in the CardsDs component

michael-odonovan:
### Why is this required?
When no link is present in the CardsDs component, the styling of the image goes astray.

### Jira ticket:
This is a bug fix that comes out of this ticket:
https://comicrelief.atlassian.net/browse/ENG-1161


michael-odonovan:
The reason there are multiple pushes is because of the mental workflow I had to do:
### To change something in a branch of the Component Library:
-pull down the Component Library to your local machine.
-make a branch.
-make the change.
-push the branch to the Component Library repo.

### Then in the contentful repo on your local machine:

#### package.json
comicrelief/component-library#your-new-branch-name
```javascriptreact
"@comicrelief/component-library": "comicrelief/component-library#cards_ds_no_link_image_styling_fix",
```
then
-delete the node_modlues/@comicrelief/component-library
then
$ npm install 
this will pull down the component library files from the new branch you have just created.

michael-odonovan:
I did a bit of a hack and just made a MediaWrapper that has the same stylings as the MediaLink (the only element in the dom that differs between a card with a link and one without). It appears in the CL in this branch. Now trying to see if the problem is fixed in CR.com. Weird that can't see the bug in the CL. Have checked and checked through CR.com and can't find anything that should be affecting these images. If this is the solution, will refactor of course.

AndyEPhipps (approved):

PR #472, opened 2022-05-16
docs: revise PR template following Eng Forum

michael-odonovan:
### What is it doing?
Adds more information about the required format of Semantic PRs.

### Why is this required?
Add some documentation after discussions about PRs in our Engineering Forum.


PULL_REQUEST_TEMPLATE.md
```diff
@@ -1,8 +1,26 @@
...
+
+#### link to Jira ticket:
+Add a link to the Jira ticket.
+
+
+## Important:
+#### !!! make sure to squash merge
+
+### Note on Commits: 
+#### Currently not essential to pass through testing pipeline, but probably good to follow Conventional Commits specs anyway. These commit messages will be squash merged in the end.
```
seb-cr, line 26: > but probably good to follow Conventional Commits specs anyway

I disagree; it doesn't really make sense for work-in-progress commits.

seb-cr (commented):
Good to get this written down. One thing missing that I think should be included is a point about tests and docs.

What do you think about having a checklist? Rough example:

- [ ] I have filled in the above sections
- [ ] My PR title follows the Conventional Commit spec
- [ ] I have added tests to cover new or changed behaviour
- [ ] I have updated any relevant documentation

PULL_REQUEST_TEMPLATE.md
michael-odonovan: Yeh I agree. Will change now.

michael-odonovan: > 

Sounds good. I'll add the checkboxes.
thanks Seb

PULL_REQUEST_TEMPLATE.md
```diff
@@ -1,8 +1,46 @@
...
+<label>I have filled out the PR description as per the template above.</label>
+<br>
+
+<input type="checkbox">
+<label>I have added tests to cover new or changed behaviour.</label>
+<br>
+
+<input type="checkbox">
+<label>I have updated any relevant documentation.</label>
+<br>
```
seb-cr, line 44: GitHub markdown supports checklists natively 🙂 

```markdown
- [ ] this is a tickbox
- [ ] here's another one
```

seb-cr: Odd, I'm 90% sure that I was able to click them in the past... Just to test I ticked a couple in my earlier comment and it seems to work? Possible that it's only allowed if you're the person who wrote the comment in the first place. It also wouldn't work in the Markdown file preview since this is not editable.

PULL_REQUEST_TEMPLATE.md
```diff
@@ -1,8 +1,30 @@
-Type: patch
+### PR Titles
+#### To pass testing pipeline, these need to follow Conventional Commits spec:
+https://www.conventionalcommits.org/en/v1.0.0/
+e.g.
+`feat: create NewForm component.`
+or
+`fix: update broken link in NewForm component.`
```
seb-cr, line 7: Just spotted one last thing, our commit message style is without a fullstop at the end. Would be good to match this in the examples here.

seb-cr (approved):
This slipped off my radar. I left a final comment, then let's get it merged!

PULL_REQUEST_TEMPLATE.md
michael-odonovan: Yeah cool, all done. Thanks Seb.

PR #473, opened 2022-05-16
fix: Reset associated MP fields when option unselected

AndyEPhipps:
Type: patch

Fixes #https://comicrelief.atlassian.net/browse/ENG-1520

- As per the ticket, when a MP option is toggled on, a value is supplied, then then toggle off again, we want to ensure we're not accidentally able to submit unvalidated values.
- Sooooo we're resetting all fields associated with the checkbox option

### DIFF
--- a/src/components/Organisms/MarketingPreferencesDS/_CheckAnswer.js
+++ b/src/components/Organisms/MarketingPreferencesDS/_CheckAnswer.js
@@ -19,6 +19,12 @@ const CheckAnswer = ({
       newVal = e.target.value === 'yes' ? e.target.value : 'no';
     } else {
       newVal = '';
+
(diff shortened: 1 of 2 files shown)


AndyEPhipps:
Fix in action:

https://user-images.githubusercontent.com/4737220/168597760-da9fb147-ccef-49ce-a494-f0cc0aaef5e0.mov

AndyEPhipps:
Testable here too: https://deploy-preview-473--cr-component-library.netlify.app/#!/MarketingPreferencesDSForm

seb-cr (approved):
As discussed: there's a slight UX dent is that pre-filled fields will be wiped if you toggle the tickboxes, but it's a bigger problem if we're submitting invalid values to the backend. And who mindlessly toggles tickboxes anyway? 😄

PR #474, opened 2022-05-19
feat: add Styled Components babel plugin for better CSS classnames

michael-odonovan:
What is it doing?
Adds babel plugin to dev dependencies so that CSS classnames make reference to the Component name, rather than just dynamically generated hash.

Why is this required?
Very hard to debug CSS style issues with just dynamically generated classnames like rh5b2$h

### DIFF
--- a/package.json
+++ b/package.json
@@ -73,6 +73,7 @@
   },
   "devDependencies": {
     "babel-eslint": "^10.0.2",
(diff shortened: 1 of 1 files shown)


michael-odonovan:
This would be a handy addition. It is already present in the CR.com repo.

PR #475, opened 2022-05-19
fix: replace yup-phone with custom validation in MPs

AndyEPhipps:
As per discussions [here](https://github.com/comicrelief/comicrelief-contentful/pull/1102), yup-phone isn't fit for purpose

### DIFF
--- a/src/components/Organisms/MarketingPreferencesDS/_MarketingPrefsConfig.js
+++ b/src/components/Organisms/MarketingPreferencesDS/_MarketingPrefsConfig.js
@@ -64,6 +64,8 @@ const buildValidationSchema = overrideOptions => {
   // Override with any custom options supplied
   const mpValidationOptions = merge(defaultOptions, overrideOptions);
 
+  const phoneRegex = /^(((((\+44)|(0044))\s?\d{4}|\(?0\d{4}\)?)\s?\d{3}\s?\d{3})|((((\+44)|(0044))\s?\d{3}|\(?0\d{3}\)?)\s?\d{3}\s?\d{4})|((((\+44)|(0044))\s?\d{2}|\(?0\d{2}\)?)\s?\d{4}\s?\d{4}))(\s?\\#(\d{4}|\d{3}))?$/;
+
   const mpValidationSchema = yup.object({
     mp_email: yup.string()
       .when('mp_permissionEmail', {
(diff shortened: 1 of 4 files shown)


AndyEPhipps:
I note that the regex isn't specifically for landlines or mobiles (so the different usage here is a little off)- however, this is a huge improvement over yup-phone (covers all of the test cases), and I aim to put together 2 more regex patterns to offer that specificity.

seb-cr:
@AndyEPhipps I wonder if we ought to break out email and phone validation into a new package. We've got different rules everywhere; `data-models` was supposed to solve it by bringing all validation into one place, but that package is necessarily private, so not suitable for `component-library` which is public. I'm thinking a new `@comicrelief/common-validation` package could be used as a dependency for both, and then we can nail down all the test cases in one place and use the same validation in every frontend and backend.

src/components/Organisms/MarketingPreferencesDS/_MarketingPrefsConfig.js
```diff
@@ -65,6 +64,8 @@ const buildValidationSchema = overrideOptions => {
   // Override with any custom options supplied
   const mpValidationOptions = merge(defaultOptions, overrideOptions);
 
+  const phoneRegex = /^(((((\+44)|(0044))\s?\d{4}|\(?0\d{4}\)?)\s?\d{3}\s?\d{3})|((((\+44)|(0044))\s?\d{3}|\(?0\d{3}\)?)\s?\d{3}\s?\d{4})|((((\+44)|(0044))\s?\d{2}|\(?0\d{2}\)?)\s?\d{4}\s?\d{4}))(\s?\\#(\d{4}|\d{3}))?$/;
```
seb-cr, line 67: This looks like it accepts only _UK_ phone numbers, is that right?

In which case, we ought to say in the validation messages `Please enter a valid UK phone number` rather than imply that international numbers are invalid.

AndyEPhipps: Yeah okay cool, although we'd know need to do the same over in the FSU 

@KrupaPammi - that okay for you?

AndyEPhipps: Thanks @KrupaPammi

seb-cr (approved):

PR #481, opened 2022-06-16
feat: Updating cookie banner as a learning exercise

AndyEPhipps:
### What is it doing?
Updating copy of the now unused cookie banner, to illustrate release process, etc.

### Why is this required?
'Cause I said so.


pauldolden (approved):
Hey Andy,

I don't see any issues with this at all. Linting looks great!

Cheers

PR #492, opened 2022-08-26
feat: add optional label and required prop to select component

pauldolden:
#### What is it doing?

- Adds the `optional` label prop to the select component.
- Adds the `required` flag to the select component.

#### Why is this required?
Allows for more clarity when select fields are not required.

### DIFF
--- a/src/components/Atoms/Select/Select.js
+++ b/src/components/Atoms/Select/Select.js
@@ -49,6 +49,7 @@ const Select = React.forwardRef(
       onChange,
       greyDescription,
       className,
+      optional,
       ...rest
     },
     ref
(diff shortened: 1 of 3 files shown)


AndyEPhipps (commented):
Could we add the required attribute too, while we're here?

See https://github.com/comicrelief/component-library/blob/master/src/components/Atoms/Input/Input.js#L97

pauldolden:
> Could we add the required attribute too, while we're here?
> 
> See https://github.com/comicrelief/component-library/blob/master/src/components/Atoms/Input/Input.js#L97

Yup, will do. Cheers Andy.

AndyEPhipps (approved):

PR #497, opened 2022-09-15
fix: fix header for in-situ usage

AndyEPhipps:
#### What is it doing?
I overlooked the fact that the CL is actually adding padding to its examples:

Screenshot 2022-09-15 at 10 30 33

So, within the real context:

#### Why is this required?
IT LOOKS BAD

Screenshot 2022-09-15 at 10 04 25
Screenshot 2022-09-15 at 10 04 31

#### link to Jira ticket:
Same ticket as before, although I'm not going to _hyperlink_ it, as it'll annoyingly close the ticket AGAIN before I get a chance to get int into CR properly - comicrelief.atlassian.net/browse/ENG-1784

### DIFF
--- a/src/components/Atoms/Logo/Logo.js
+++ b/src/components/Atoms/Logo/Logo.js
@@ -68,7 +68,7 @@ Logo.propTypes = {
 Logo.defaultProps = {
   rotate: false,
   sizeSm: '51px', // - to work with the header 75px height and 12px padding
-  sizeMd: '75px',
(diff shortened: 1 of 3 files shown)


AndyEPhipps:
With changes (with the CL's added padding removed via the inspector):

Screenshot 2022-09-15 at 10 49 26
Screenshot 2022-09-15 at 10 49 32

michael-odonovan (approved):

PR #505, opened 2022-10-05
fix: Donate widget updates 3

AndyEPhipps:
- Update default copy for pop up - https://comicrelief.atlassian.net/browse/ENG-1857
- Fix transparency background colour; only adding opacity when a non-transparent bg colour is chosen

To follow:
- ~~Mock ups on edits to pop up message, will feed into new styles~~ - Will keep this a separate PR, for sanity

### DIFF
--- a/src/components/Organisms/Donate/Donate.js
+++ b/src/components/Organisms/Donate/Donate.js
@@ -148,7 +148,7 @@ Donate.defaultProps = {
     'will help us fund amazing projects in the UK and around the world.',
   subtitle: '',
   noMoneyBuys: false,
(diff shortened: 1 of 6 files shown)


AndyEPhipps:
Doublechecked the value we need here; it comes out of the CMS as 'Transparent' but the snakecase function makes it 'transparent'; have added this to the relevant example, and also to the theme colourlist, so it can pass the correct CSS value [here](https://github.com/comicrelief/component-library/blob/fba80c38101d3519c94308fb1b9674fe38293723/src/components/Organisms/Donate/Donate.style.js#L22).

michael-odonovan (approved):

PR #506, opened 2022-10-06
feat: Donate widget updates 4

michael-odonovan:
#### What is it doing?
More updates to Donate widget on CR.com

#### Why is this required?
FR

#### link to Jira ticket:
https://comicrelief.atlassian.net/browse/ENG-1810

### DIFF
--- a/src/components/Organisms/Donate/Donate.js
+++ b/src/components/Organisms/Donate/Donate.js
@@ -68,7 +68,7 @@ const Donate = ({
         />
       ) : null}
 
-      <Wrapper formAlignRight={formAlignRight}>
+      <Wrapper formAlignRight={formAlignRight} aria-live="polite">
         <Header formAlignRight={formAlignRight}>
           <HeaderInner>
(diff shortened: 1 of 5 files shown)


michael-odonovan:
Info for QA / Testing:
Changes are concerned with the PopUp text box that attempts to persuade users to donate monthly rather than single donations. 
The tickets that affect actual functionality are:

**Making it so the PopUp only fires when users move away from monthly to single, using the single button:**
https://comicrelief.atlassian.net/browse/ENG-1802

**Transition effect / animation when PopUp box is closed with the "X" in the top corner:**
https://comicrelief.atlassian.net/browse/ENG-1856

All other new changes are color / style / minor cms functionality changes.

KrupaPammi:
Thanks for the info @michael-odonovan 
I tested the donate widget, I see that the popup text appears only when I click monthly and then click single button, doesn’t it appear by default when the page is loaded?

KrupaPammi (approved):

PR #516, opened 2022-10-19
feat: add chevrons to mobile view as well 

michael-odonovan:
#### What is it doing?
Add chevrons to mobile view.

#### Why is this required?
Got wires crossed and thought they were only for dekstop view.

#### link to Jira ticket:
https://comicrelief.atlassian.net/browse/ENG-1740

### DIFF
--- a/src/components/Organisms/Header/Nav/Nav.js
+++ b/src/components/Organisms/Header/Nav/Nav.js
@@ -100,6 +100,13 @@ const MainNav = ({ navItems }) => {
                     role="button"
                   >
                     {thisFirstChild.title}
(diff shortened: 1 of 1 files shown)


pauldolden (commented):
Looks fine to me.

Only question I have, (and I suppose this goes for the desktop view also, which I probably should have raised at the time), would be whether we should rotate the chevron when the sub-menu is open?

This may or may not have already been scoped out in the requirements, so if I'm raising an already discussed point, then just let me know and I'm happy to just approve.

pauldolden (approved):

PR #522, opened 2022-10-25
feat: ESU updates

pauldolden:
#### What is it doing?

- Migration of the ESU form to `react-hook-form`
- Simplification of form logic (removed unused schools form)
- Added new form fields for `firstName` and `lastName`
- Refactored logic to work with new form fields
- Updated tests to reflect new form behaviour
- Added confetti component to form success[^1]
- Removed `HeaderEsuWithIcon` component[^2]

Anything I've missed let me know 😄.

[^1]: This is a slight reworking of the Confetti component used on FSU and works quite nicely, could perhaps be refactored into a CL `Atom` for future reuse.

[^2]: This will be reimplemented inside `comicrelief-contentful` as it is more suited to being there as is quite specific, and not really open for resuse.

#### Why is this required?

There was a desire to capture more information as part of the ESU form, however the currently implementation wasn't really designed in such a way as to be extensible, and also had not received significant updates since we adopted the use of RHF. So I've taken the opportunity to update, clean up and migrate to RHF.

#### link to Jira ticket:

[ENG-1894]

[ENG-1894]: https://comicrelief.atlassian.net/browse/ENG-1894?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/components/Organisms/EmailSignUp/_Confetti.js
```diff
@@ -0,0 +1,106 @@
...
+        // This gracefully ends the animation
+        pauseAnimation();
+      }, duration);
+    }
+    return () => {
+      if (timeOut) {
+        // this clears up the animation
+        stopAnimation();
+      }
+    }; // eslint-disable-next-line react-hooks/exhaustive-deps
```
AndyEPhipps, line 89: Can we not pass in whatever dep eslint is asking for here?

AndyEPhipps (approved):

src/components/Organisms/EmailSignUp/_Confetti.js
pauldolden: I tried that initially, and it causes an infinite rendering loop. I don't actually mind those things being stale as they are coming the the dependency anyway, so I don't think this will be a problem.

pauldolden:
Thanks for approving @AndyEPhipps, I've made a couple more commits, just want to make sure they are all OK before I merge:

- Added Button Colour Prop
- Added Background Colour Prop
- Freeze the `ESU_FIELDS` object[^1]

Also added an empty commit with the `BREAKING CHANGE` label. On the of change my muscle memory squashes immediately after merging. I believe this will be picked up.


[^1]: As this is being exposed the the consumer of the component, it would be theoretically possible to break the form by removing properties from the object. This is unlikely but I feel it's worthwhile to prevent this.

AndyEPhipps (approved):
All makes sense to me, matey; assign me once it's insitu in a CR PR and I'll try and break it for real

pauldolden:
> All makes sense to me, matey; assign me once it's insitu in a CR PR and I'll try and break it for real

Will do thanks!

PR #525, opened 2022-10-31
fix: fix 'repeated ID' Donation widget giving-type buttons issue

AndyEPhipps:
#### What is it doing?
Just passing in the unique Donate row ID (from the CMS) to form dynamic IDs per row, meaning any label correlates with the right _unique_ button on the page. Have also broken out GivingType styles into their own file for cleanliness.

#### Why is this required?
As per the ticket, a second (or more; visible in the CL itself) Donate widget interaction would be changing the giving type for the _first_ widget on the page, since we had repeated IDs.

#### link to Jira ticket:
[Add a link to the Jira ticket.](https://comicrelief.atlassian.net/browse/ENG-1906)

### Quick Checklist:


AndyEPhipps:
I'm also going to add in the updates outline here:

https://comicrelief.atlassian.net/browse/ENG-1904

michael-odonovan (approved):

PR #527, opened 2022-10-31
fix: increase limit to £25000 per Donate platform, move error msg 

AndyEPhipps:
#### What is it doing?
Updates the limit to the 25000 figure as per the Donate platform itself, moves the error message, and updates tests accordingly (using newly added class to avoid brittle selectors)

#### Why is this required?
Because Curtis said he would "skkkkweeeam and skkkkweeeam" until he was sick if I didn't do it

#### link to Jira ticket:
https://comicrelief.atlassian.net/browse/ENG-1904


pauldolden (approved):
All looks good to me.

PR #530, opened 2022-11-22
feat: Make Confetti Separate Component

michael-odonovan:
### PR description
Make Confetti Separate Component

#### Why is this required?
We are using this Confetti idea in multiple places / DRY

#### Jira Ticket
https://comicrelief.atlassian.net/browse/ENG-1959


michael-odonovan:
This confetti runs as soon as you run the local server, so is kind of annoying / not the best way of viewing the actual confetti.

AndyEPhipps (changes requested):
Looking good, but could we remove it from the ESU component, and add this version in?

michael-odonovan:
Yeh we have a ticket for that, which I am gonna handle after this sprint. One step ahead of you...

michael-odonovan:
I will do the other FSU usage at the same time as well

AndyEPhipps:
Any reason why we can't just do it now as part of this work?

AndyEPhipps:
Eh, screw it, I guess we can do it later.

Have you been able to test within your CRcom build locally?

AndyEPhipps (approved):
Nice oooone 👍

PR #539, opened 2022-12-20
feat: create CMS field to override "other amount" money buy message

michael-odonovan:
PR description
What is it doing?
Create CMS field to override "other amount" money buy message.

Why is this required?
Donate improvements

link to Jira ticket:
https://comicrelief.atlassian.net/browse/ENG-1984

### Quick Checklist:

### DIFF
--- a/src/components/Organisms/Donate/Donate.js
+++ b/src/components/Organisms/Donate/Donate.js
@@ -36,7 +36,8 @@ const Donate = ({
   noMoneyBuys,
   PopUpText,
   chooseAmountText,
-  isDesktopOverride
+  isDesktopOverride,
+  moneyBuyCustomMessage
 }) => {
   let isDesktop = useMediaQuery({ query: `(min-width: ${screen.medium})` });
 
@@ -113,6 +114,7 @@ const Donate = ({
           noMoneyBuys={noMoneyBuys}
           PopUpText={PopUpText}
           chooseAmountText={chooseAmountText}
+          moneyBuyCustomMessage={moneyBuyCustomMessage}
         />
       </Wrapper>
     </Container>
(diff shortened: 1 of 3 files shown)


AndyEPhipps (changes requested):
Hey man, can you add an example to the CL that does pass a custom message, to show everything working nicely?

AndyEPhipps:
Hey man, it looks like you need to add in a check; this custom value is being added to _all_ contexts here, when it should just be for an 'other amount':

AndyEPhipps (changes requested):
Hey man, it looks like you need to add in a check; this custom value is being added to _all_ contexts here, when it should just be for an 'other amount':

Screenshot 2023-01-09 at 09 25 24

michael-odonovan:
@AndyEPhipps 
Yeh I thought the custom message was supposed to override the lot, however have checked with Rosie and you are right, it only applies when you input an ammount. Have changed now.

AndyEPhipps:
Hey man, now I zoom out a bit , I think we've got the wrong end of the stick here. 

The pre-existing `otherDescription` prop that does exactly what we need already, we're currently just overriding the default prop value with a non-dynamic value in the build [here](https://github.com/comicrelief/comicrelief-contentful/blob/master/src/components/Donate/Donate.js#L103). 

_This_ work is to add a field to the CMS that is passed to override _that_ otherDescription prop, meaning that text value can be set on a widget-by-widget basis; sorry for not catching this earlier

michael-odonovan:
Ok that's cool, I'll have another look at it. I checked out "otherDescription" prop ages ago but thought it didn't quite do what was required, however I'll go back to CMS and check it all out again.

PR #542, opened 2023-01-16
feat: updates to `Donate` component to allow for more advanced theming

pauldolden:
#### What is it doing?
- Adds textColor field.
- Adds overlayColor field (for desktop overlay colour).
- Adds a buttonColor field to allow for future theming of the submit button.
- Changes backgroundColour to refer to mobile/tablet background colour.
- Updates snapshots.

#### Why is this required?
The theming work for the Donate widget requires some updates and changes to allow for some more fine-grained theming of the widget on both desktop and mobile/tablet views.

As per Jira: [ENG-2069]

### Quick Checklist:

[ENG-2069]: https://comicrelief.atlassian.net/browse/ENG-2069?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/components/Organisms/Donate/Donate.js
+++ b/src/components/Organisms/Donate/Donate.js
@@ -19,6 +19,7 @@ const Donate = ({
   cartID,
   clientID,
   backgroundColor,
+  buttonColor,
   donateLink,
   data,
   title,
@@ -113,6 +114,7 @@ const Donate = ({
           noMoneyBuys={noMoneyBuys}
           PopUpText={PopUpText}
           chooseAmountText={chooseAmountText}
+          buttonColor={buttonColor}
         />
       </Wrapper>
     </Container>
@@ -137,6 +139,7 @@ Donate.propTypes = {
   mobileImages: PropTypes.string,
   mobileAlt: PropTypes.string,
   backgroundColor: PropTypes.string,
+  buttonColor: PropTypes.string,
   mbshipID: PropTypes.string,
   noMoneyBuys: PropTypes.bool,
   PopUpText: PropTypes.string,
@@ -156,6 +159,7 @@ Donate.defaultProps = {
   mobileImages: null,
   mobileAlt: '',
   backgroundColor: 'blue_dark',
+  buttonColor: 'red',
   mbshipID: null,
   otherAmountText:
     'will help us fund amazing projects in the UK and around the world.',
--- a/src/components/Organisms/Donate/Donate.js
+++ b/src/components/Organisms/Donate/Donate.js
@@ -20,6 +20,7 @@ const Donate = ({
   clientID,
   backgroundColor,
   buttonColor,
+  textColor,
   donateLink,
   data,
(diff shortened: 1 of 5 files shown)


AndyEPhipps:
Screenshot 2023-01-17 at 16 07 29


Would be good to get bg colours added to these, and maybe put in a suitable 'using default transparent colour' example. I know it's a bit unwieldy with so many options, but good to have such things there to catch any weird bugs upd

pauldolden:
> Screenshot 2023-01-17 at 16 07 29
> 
> Would be good to get bg colours added to these, and maybe put in a suitable 'using default transparent colour' example. I know it's a bit unwieldy with so many options, but good to have such things there to catch any weird bugs upd

Cool. thanks Andy. I'll do that.

AndyEPhipps:
Given not about to use this in another context, I guess it does makes sense to do the 'theme' logic/mapping in the CRcom rather than here?

The only benefit doing it here would mean we could easily have a handful of working examples to see how all these options mesh; maybe it's just a case of making sure your CR PR env (and ultimately master_model) has a comprehensive demo page for this component to 'demo' them all?

pauldolden:
> Given we're not about to use this in _another_ context, I guess it does makes sense to do the 'theme' logic/mapping in the CRcom rather than here?
> 
> The only benefit doing it here would mean we could easily have a handful of working examples to see how all these options mesh; maybe it's just a case of making sure your CR PR env (and ultimately master_model) has a comprehensive demo page for this component to 'demo' them all? What do you reckon?

Yeah I was unsure about this myself, there is definitely some logic in the CR.com part of this PR, that I'm not a huge fan of, and does need a bit of refactoring, and it might be plausible to clean things up by doing it here. 

I think logic _should_ belong in CR.com, but I couldn't give you an iron clad reason why I believe that to be the case. One thing I do think might be a good compromise, would be to clean up the prop names a little to make it a bit clearer what things like `backgroundColor` actually mean.

AndyEPhipps (changes requested):
Leaving an additional comment so GH gets off my back

pauldolden:
> Thinking... I suppose keeping it this way would mean all new 'theme' work for this could live purely within a CRcom PR, rather than having to do both? New 'theme' option added to the CMS, handled and mapped appropriately in the CRcom codebase, _badda-bing_.
> 
> Rather than over-engineer this component (when it should perhaps be a _bit_ 'dumb'), maybe we **are** best to keep the 'theme' definition and colour mapping within CRcom. If we _do_ ever end up using this component in another build, I'll do the legwork _and_ eat my hat

Yeah I think that's probably the right call. I'll get these quick changes made and hopefully get this merged, then finish off the CR.com stuff.

AndyEPhipps (approved):

AndyEPhipps:
💩

PR #548, opened 2023-02-09
fix: make MP field values persist when hidden

AndyEPhipps:
#### What is it doing?
Adds in a check so that we're only clearing out MP field values when it's a scenario when the user is being shown the field, not a "pass the value behind the scenes, keep the field hidden" scenario.

#### Why is this required?
Bug raised

#### link to Jira ticket:
[Link to the Jira ticket.](https://comicrelief.atlassian.net/browse/ENG-2191)

### Quick Checklist:


seb-cr (commented):
Thanks @AndyEPhipps, looks good.

Is there somewhere here that we can add a regression test for this? I feel like something similar happened before, and test cases are the easiest way to avoid subtle bugs creeping in during refactors.

seb-cr (approved):

PR #552, opened 2023-02-22
feat: allow Donate row to show decimals properly

AndyEPhipps:
#### What is it doing?
Allow Donates row to show decimals properly

#### link to Jira ticket:
[Add a link to the Jira ticket.](https://comicrelief.atlassian.net/browse/ENG-2250)

### Quick Checklist:

### DIFF
--- a/src/components/Organisms/Donate/Form/Form.js
+++ b/src/components/Organisms/Donate/Form/Form.js
@@ -164,7 +164,7 @@ const Signup = ({
                 <MoneyBox
                   isSelected={amountDonate === value}
                   amount={value}
-                  description={`£${value}`}
+                  description={`£${amountFormatter(value)}`}
                   setOtherAmount={() => setAmountDonate(parseFloat(value))}
                   key={value}
                   name={`${mbshipID}--moneyBuy${index + 1}`}
--- a/src/components/Organisms/Donate/Form/Form.js
+++ b/src/components/Organisms/Donate/Form/Form.js
@@ -39,7 +39,7 @@ const Signup = ({
 }) => {
   const [givingType, setGivingType] = useState('single');
   const [errorMsg, setErrorMsg] = useState(false);
-  const [amountDonate, setAmountDonate] = useState(' ');
+  const [amountDonate, setAmountDonate] = useState(null);
   const [moneyBuyCopy, setMoneyBuyCopy] = useState(true);
   const [popOpen, setPopOpen] = useState(false);
   // In order to keep track of whether the user has ever been shown the popup
--- a/src/components/Organisms/Donate/Form/Form.js
+++ b/src/components/Organisms/Donate/Form/Form.js
@@ -3,7 +3,7 @@ import PropTypes from 'prop-types';
 
 import PopUpComponent from './PopUpComponent';
 import Text from '../../../Atoms/Text/Text';
-import MoneyBox from '../MoneyBox/MoneyBox';
+import MoneyBuy from '../MoneyBuy/MoneyBuy';
 import {
   handleDonateSubmission,
   isAmountValid,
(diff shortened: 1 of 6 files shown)


src/components/Organisms/Donate/dev-data/data.js
```diff
@@ -5,7 +5,7 @@ export default {
       {
         description:
           'a pregnant woman in Uganda to have access to a birth kit so she can have a clean, safe delivery.',
-        value: 10
+        value: 10.1
```
AndyEPhipps, line 8: Added this in to verify how we're transforming the data exactly as it comes out of the CMS.

src/components/Organisms/Donate/__snapshots__/Donate.test.js.snap
```diff
@@ -1531,7 +1531,7 @@ exports[`Single donation renders correctly 1`] = `
                   required={false}
                   step="0.01"
                   type="number"
-                  value={20}
```
AndyEPhipps, line 1534: Surprised this ever worked tbh

src/utils/Membership.js
```diff
@@ -5,6 +5,13 @@ const onKeyPress = event => {
   if (/\+|-|e/.test(keyValue)) event.preventDefault();
 };
 
+const amountFormatter = amount => {
```
pauldolden, line 8: Tangentially related to the work here, but initially found this confusing because this is not strictly speaking `Membership` specific logic. Obviously, small quibble but maybe this function belongs elsewhere or the filename should be changed.

src/components/Organisms/Donate/MoneyBox/MoneyBox.js
```diff
@@ -50,23 +54,27 @@ const MoneyBuy = ({
...
-  />
-);
+}) => {
+  // Set decimal points accordingly
+  const formattedAmount = amountFormatter(amount);
+  return (
+    <MoneyBox
+      {...rest}
+      aria-label={description}
+      value={`${currency} ${formattedAmount}`}
```
pauldolden, line 64: More of a query than anything else. This component accepts a currency, and an amount, but we just pass both as single string. I just wondered why this was? Potentially again, outside of the scope of this work but just something I noticed and thought I would query.

src/components/Organisms/Donate/Form/Form.js
```diff
@@ -163,7 +164,7 @@ const Signup = ({
                 <MoneyBox
                   isSelected={amountDonate === value}
                   amount={value}
```
pauldolden, line 166: Again, nothing major just a query. In the other instances we pass the `amountFormatter(value)` into the component, but here we handle this inside the component.

```js
const formattedAmount = amountFormatter(amount);
```

Was there any specific reason for this? Again just largely unimportant point.

pauldolden (commented):
Everything looks good, and all works fine. Nice work.

Only things I picked up on were some of my own curiosities.

src/utils/Membership.js
AndyEPhipps: Yeah, I guess these shared util functions should be broken-out into something more generic-sounding; will tidy this up in another PR if that's cool

src/components/Organisms/Donate/MoneyBox/MoneyBox.js
AndyEPhipps: Ehhh, there's definitely a lot to be refactored in this old (not my) work; I guess there's cause to keep both those variables separate to do _number_-specific things to `amount`, but we need the currency symbol here for aesthetics purely?

Not sure if I misunderstood you here?

pauldolden: That makes sense, it works fine this way, was just curious as it its purpose. 😄

src/components/Organisms/Donate/Form/Form.js
AndyEPhipps: Ahhh nah, I think this was left over from me figuring out stuff, good catch. I'll move it out, and clear up the naming in this MoneyBuy/Moneybox thing, it just confused tf out of me

pauldolden: Yeah, I found that confusing as well.

pauldolden (approved):
🚀  All good.

MoneyBuyButton makes _so_ much more sense than MoneyBox 🐷.

PR #557, opened 2023-04-05
feat: donate button show amount

michael-odonovan:
#### What is it doing?
Adds the amount to be donated to the button when the moneybuys or custom amount are inputed / selected.

#### Why is this required?
FR / BBC around RND23

#### link to Jira ticket:
https://comicrelief.atlassian.net/browse/ENG-2353

### Quick Checklist:


AndyEPhipps (changes requested):
Looks good to me, man- just two things:

While we're here, do you think we could update the button font to `inherit` the proper font-family?

Also, can you see what Curtis reckons of this? I feel like there's a nicer message to show here rather than just omitting the value:

Screenshot 2023-04-06 at 14 52 14

michael-odonovan:
ok cool, will do

michael-odonovan:
- Curtis suggested that when the input field is left blank / cleared, the click outside should return the money buy to the default.
- I also removed the lonely "£" when no value is entered.

src/components/Organisms/Donate/Form/useClickOutside.js
```diff
@@ -0,0 +1,41 @@
...
+        setAmountDonate(parseFloat(thisAmount));
+      }
+    }
+    // Bind the event listener
+    document.addEventListener('mousedown', handleClickOutside);
+    return () => {
+      // Unbind the event listener on clean up
+      document.removeEventListener('mousedown', handleClickOutside);
+    };
+  }, [amountRef, errorMsg]);
```
AndyEPhipps, line 40: Can you add these useEffect deps in?

src/components/Organisms/Donate/Form/useClickOutside.js
```diff
@@ -0,0 +1,41 @@
...
+  givingType,
+  singleGiving,
+  errorMsg,
+  regularGiving
+) {
+  useEffect(() => {
+    /**
+     * Alert if clicked on outside of element
+     */
+    function handleClickOutside(event) {
```
AndyEPhipps, line 16: Hmm could we move the function declaration out of the useEffect so we're not recreating  it unnecessarily each time?

src/components/Organisms/Donate/Form/useClickOutside.js
AndyEPhipps: Hmmm if you're binding the function to the mouseup and down functions, why does it need to be in a useEffect, retriggering when amountRef or errorMsg gets updated?

src/components/Organisms/Donate/Form/useClickOutside.js
```diff
@@ -0,0 +1,41 @@
...
+          : givingData.moneybuys[0].value;
+
+        setAmountDonate(parseFloat(thisAmount));
+      }
+    }
+    // Bind the event listener
+    document.addEventListener('mousedown', handleClickOutside);
+    return () => {
+      // Unbind the event listener on clean up
+      document.removeEventListener('mousedown', handleClickOutside);
```
AndyEPhipps, line 38: V nice

michael-odonovan:
@AndyEPhipps 
Fiddled around with this a lot yesterday. 
- have moved the code back into the main Form.js file.
- the latest commit has the handleClickOutside function outside of the useEffect, however it will complain / error unless you wrap it in the useCallback hook.
- the previous commit has the handleClickOutside function inside the useEffect, and this obvs also works.
I'll leave it to you to decide which way to go.
I also added better button colours to the markdown examples as I couldn't see the red writing with the red background (this is on master as well).

AndyEPhipps:
Assigning Paul to review this too, so I'm not blocking it while I'm tied-up

michael-odonovan:
ok cool

pauldolden:
Everything seems to work here as expected here. Only question is about the snapshots, they appear to render the null values. Nothing that appears to affect the behaviour, but worth mentioning.

michael-odonovan:
> Only question is about the snapshots, they appear to render the null values. 

Cheers Paul. However I don't understand this bit. I can see src/components/Organisms/Donate/__snapshots__/Donate.test.js.snap, the jsx being changed by Jest/snapshots. But I don't understand null values?

pauldolden:
> > Only question is about the snapshots, they appear to render the null values.
> 
> Cheers Paul. However I don't understand this bit. I can see src/components/Organisms/Donate/**snapshots**/Donate.test.js.snap, the jsx being changed by Jest/snapshots. But I don't understand null values?

Hey Mike, apologies for the confusion here:
```
Donate £  now
```

I just meant that these are being rendered as per the above in the snapshots, just a comment more than anything else.

michael-odonovan:
oh yeah gotcha now => I suppose Jest can't parse the Js conditionals and just grabs from a state somewhere along the line?
I recon prob not worth messing with it as we just use snaps to check things haven't changed suddenly.
What you recon?

michael-odonovan:
Have changed null to 10 in the initial useState for the donation value, which has sorted this issue.

pauldolden:
> Have changed null to 10 in the initial useState for the donation value, which has sorted this issue.

Cool, I'm happy to approve this. All seems to work as expected. Unless @AndyEPhipps and @KrupaPammi have anything else to add?

michael-odonovan:
Have sorted out Button issue. Error message is in view.

AndyEPhipps:
> Tried to use setFocus() on the input after clicking the button, but something isn't working, so looks quite tricky to get that

I'll give this a look as part of the re-review, hold on your hat

AndyEPhipps (approved):
Good stuff, mate 👍

AndyEPhipps (changes requested):
Although it looks like you've remove recent work with that master merge; can re-do this please

PR #565, opened 2023-05-13
refactor: using depcheck to find and then remove unused dependencies

michael-odonovan:
#### What is it doing?
Using depcheck to find and then remove unused dependencies:
https://github.com/depcheck/depcheck

#### Why is this required?
BAU - Bug Fixes

#### link to Jira ticket:
https://comicrelief.atlassian.net/browse/ENG-2522

### Quick Checklist: (Component Library)

- [X] My PR title follows the Conventional Commit spec.

- [X] I have filled out the PR description as per the template above.

- [X] I have added tests to cover new or changed behaviour.

- [X] I have updated any relevant documentation.


package.json
```diff
@@ -73,7 +73,6 @@
     ]
   },
   "devDependencies": {
-    "babel-eslint": "^10.0.2",
```
AndyEPhipps, line 77: Hmmm, You can check to see how it's been originally added by using Blame: 

https://github.com/comicrelief/component-library/commit/e124d768104613c2ed7681eb9272aa8bac3810ef

Even after removing it, it's still listed as a dependency of react-scripts, so gets installed regardless:

https://github.com/comicrelief/component-library/blob/8c0f2b8bb16a4f4ee8ac135669af06733f5f760f/yarn.lock#L10430

It's possible it said package didn't **use** to have it as an explicit dep or something, but it does now, so we don't need to have it locally? 

Shrug, don't know. I can see this builds both the dev and prod/build version fine, and the pipeline is happy, so I'm happy for this to be removed for now; we can always come back to it if needs be.

Good stuff, mate; very methodical 👍 §

AndyEPhipps (approved):

PR #569, opened 2023-05-21
fix: widget decimal bug 2536

michael-odonovan:
#### What is it doing?
Came up as part of Krupa testing this ticket:
https://comicrelief.atlassian.net/browse/ENG-2353

#### Why is this required?
bugfix for better UX

#### link to Jira ticket:
https://comicrelief.atlassian.net/browse/ENG-2536

### Quick Checklist:

### DIFF
--- a/src/components/Organisms/Donate/Form/Form.js
+++ b/src/components/Organisms/Donate/Form/Form.js
@@ -257,7 +257,7 @@ const Signup = ({
               value={amountDonate}
               pattern="[^[0-9]+([,.][0-9]+)?$]" // this only applies on submit
               placeholder="0.00"
-              onChange={e => setAmountDonate(e.target.value)}
+              onChange={e => setAmountDonate(e.target.value.trim())}
               aria-label="Input a different amount"
               ref={amountRef}
             />


AndyEPhipps (changes requested):
> The decimal example from the moneyBuys was removed as it would be counter-productive for us to ask for decimal amounts in moneybuys, and it too was creating issues (that won't be mirrored in prod).

I _largely_ hear you on this front, **but** this work was done in order to address a realworld example for a smaller moneybuy amount: https://github.com/comicrelief/component-library/pull/552, so I'd prefer we add it back in to show the code doing exactly what it should in these kinda scenarios.

If this fix makes this impossible, we'll need to suss out how to handle it :/

src/components/Organisms/Donate/Form/Form.js
```diff
@@ -52,7 +52,7 @@ const Signup = ({
     // If a specific 'other amount' has been passed down, use it,
     // otherwise assign based on the associated moneybuys:
     if (otherAmountValue) {
-      setAmountDonate(parseFloat(otherAmountValue));
+      setAmountDonate(otherAmountValue);
```
AndyEPhipps, line 55: Hm, why did you need to remove this parse function?

AndyEPhipps (changes requested):
So now it's a string field, I'm able to enter spaces either side of the value, and the button ends up showing it like `£ value`:

Screenshot 2023-06-08 at 12 21 46

It's not causing any _obvious_ subsequent problems within Donate itself (I can see the right value stored and used), but it does makes the forwarding URL look like:

donation.comicrelief.com/?clientOverride=donate&**amount=%20%20%20%20%20123.45%20%20%20%20%20%20%20**&currency=GBP&givingType=single&cartId=default-comicrelief&affiliate=generic&siteurl=https://deploy-preview-569--cr-component-library.netlify.app/&rowID=mship-1&popUpShown=false

Which may introduce more problems and also nause up some Google Analytics stuff.

Do you think you could either update the input field to ignore any spaces being inputted (I guess you could just strip it when saving to state), or just strip the value before it gets used in the forwarding URL?

PR #572, opened 2023-06-05
feat: optional Title field for Donate widget

AndyEPhipps:
#### What is it doing?
Makes the Donate widget paragraph totally optional, only rendering markup when appropriate 

#### Why is this required?
More flexibility

#### link to Jira ticket:
https://comicrelief.atlassian.net/browse/ENG-2131

### Quick Checklist:


src/components/Organisms/Donate/Donate.js
```diff
@@ -94,6 +94,7 @@ const Donate = ({
           <HeaderInner>
             {subtitle && (
               <>
+                {title && (
```
AndyEPhipps, line 97: This is still added _within_ the Subtitle check as I'm 99% sure it was decided we'd never need a title without a subtitle.

I think we possibly used to use `Title` as the 'listing' name in the CMS (hence in being required), but there's now a required `Name` field, which is purely for this CMS admin.

pauldolden (approved):

AndyEPhipps:
> This is absolutely fine, and I am happy to approve.
> 
> General talking point about short-circuit conditional rendering in React though, my personal preference generally would be to not conditionally render like this with strings, as although in most cases it works perfectly fine, in some instances it can cause some weird UI errors, that are annoying to debug. Generally, I find it's better to cast strings to booleans if we are using them in this way:
> 
> ```js
> {!!subtitle && (
>    <>
>  {Boolean(title) && (
> ```

Ah yeah, that is honestly way better; I was, for my sins, a bit on autopilot, just rehashing what was there already.

I'm gonna update this to use the Boolean casting as it's less ALARMING than the double-exclaim, and also clearer as to what it's doing.

PR #588, opened 2023-09-27
fix: Fix Cypress

AndyEPhipps:
#### What is it doing?
Addresses Cypress failures that arrived out of nowhere (possibly underlying Chrome engine updates happening in the background), which are blocking us from releasing any new versions of the CL.

The PR updates Cypress and also updates the tests to suit the new requirements.

#### Why is this required?
CL BROKED

#### link to Jira ticket:
[Add a link to the Jira ticket.](https://comicrelief.atlassian.net/browse/ENG-2744)

### Quick Checklist:


cypress/e2e/components/atoms/button.cy.js
```diff
@@ -102,12 +98,5 @@ describe('Button component', () => {
         .find('textarea')
         .should('exist');
     });
-
```
AndyEPhipps, line 105: For whatever reason, these steps were causing tests to fail, even though the functionality works fine locally.

Given that this is testing CL itself, rather than our code, it seemed like a waste of time and energy to debug this right now, so I pulled out every instance of it across our tests.

cypress/e2e/components/atoms/button.cy.js
```diff
@@ -63,17 +65,11 @@ describe('Button component', () => {
...
-    it('closes code on click', () => {
```
AndyEPhipps, line 67: For whatever reason, these steps were causing tests to fail, even though the functionality works fine locally.

Given that this is testing CL itself, rather than our code, it seemed like a waste of time and energy to debug this right now, so I pulled out every instance of it across our tests.

KrupaPammi (approved):

PR #590, opened 2023-10-02
feat: create postcode lookup component in Component Library

michael-odonovan:
#### What is it doing?
Creates new postcode lookup component in our Component Library that can be used in forms that need a postcode lookup.

#### Why is this required?
DRY, building up on Component Library.

#### link to Jira ticket:
https://comicrelief.atlassian.net/browse/ENG-2665

### Quick Checklist:

### DIFF
--- a/src/components/Molecules/Lookup/Lookup.js
+++ b/src/components/Molecules/Lookup/Lookup.js
@@ -5,17 +5,21 @@ import TextInputWithDropdown from '../../Atoms/TextInputWithDropdown/TextInputWi
 import spacing from '../../../theme/shared/spacing';
 import ButtonWithStates from '../../Atoms/ButtonWithStates/ButtonWithStates';
 
-const Button = styled(ButtonWithStates)`${({ theme }) => css`
+const Button = styled(ButtonWithStates)`${({
+  theme,
+  buttonTextColour,
+  buttonBgColour,
+  buttonTextHoverColour,
+  buttonHoverBgColour
+}) => css`
   margin-top: ${spacing('m')};
-  color: ${theme.color('grey_dark')};
-  border: 2px solid ${theme.color('grey_dark')};
-  background-color: ${theme.color('white')};
-  padding-left: ${spacing('lg')};
-  padding-right: ${spacing('lg')};
+  color: ${buttonTextColour || theme.color('white')};
+  background-color: ${buttonBgColour || theme.color('red')};
   &:hover {
-    color: ${theme.color('grey_dark')};
-    background-color: ${theme.color('white')};
+    color: ${buttonTextHoverColour || theme.color('white')};
+    background-color: ${buttonHoverBgColour || theme.color('red_dark')};
   }
+  padding: 0 ${spacing('lg')};
 `}`;
 
 const KEY_CODE_ENTER = 13;
@@ -25,6 +29,10 @@ const KEY_CODE_ENTER = 13;
  * @param label
  * @param placeholder
  * @param buttonText
+ * @param buttonTextColour
+ * @param buttonBgColour
+ * @param buttonHoverBgColour
+ * @param buttonHoverTextColour
  * @param lookupHandler
  * @param mapOptionToString
  * @param onSelect
@@ -39,6 +47,10 @@ const Lookup = ({
   label,
   placeholder,
   buttonText,
+  buttonTextColour,
+  buttonBgColour,
+  buttonHoverBgColour,
+  buttonHoverTextColour,
   lookupHandler,
   mapOptionToString,
   onSelect,
@@ -106,6 +118,10 @@ const Lookup = ({
         loading={isSearching}
         disabled={isSearching}
         loadingText="Searching"
+        buttonTextColour={buttonTextColour}
+        buttonBgColour={buttonBgColour}
+        buttonHoverBgColour={buttonHoverBgColour}
+        buttonHoverTextColour={buttonHoverTextColour}
       >
         {buttonText}
       </Button>
@@ -118,6 +134,10 @@ Lookup.propTypes = {
   label: PropTypes.string.isRequired,
   placeholder: PropTypes.string.isRequired,
   buttonText: PropTypes.string.isRequired,
+  buttonTextColour: PropTypes.string,
+  buttonBgColour: PropTypes.string,
+  buttonHoverBgColour: PropTypes.string,
+  buttonHoverTextColour: PropTypes.string,
   lookupHandler: PropTypes.func.isRequired,
   mapOptionToString: PropTypes.func.isRequired,
   onSelect: PropTypes.func.isRequired,
@@ -127,7 +147,11 @@ Lookup.propTypes = {
 
 Lookup.defaultProps = {
   noResultsMessage: 'Sorry, could not find any results for your search',
-  dropdownInstruction: ''
+  dropdownInstruction: '',
+  buttonTextColour: 'white',
+  buttonBgColour: '#E52630',
+  buttonHoverTextColour: 'white',
+  buttonHoverBgColour: '#890B11'
 };
 
 export default Lookup;
--- a/src/components/Molecules/Lookup/Lookup.md
+++ b/src/components/Molecules/Lookup/Lookup.md
@@ -55,6 +55,10 @@ const schoolToString = school => `${school.name}, ${school.post_code}`;
     lookupHandler={schoolFetcher}
     mapOptionToString={schoolToString}
     onSelect={address => alert(JSON.stringify(address, null, 2))}
+    // buttonTextColour='black'
+    // buttonBgColour='#f04257'
+    // buttonHoverTextColour='white'
+    // buttonHoverBgColour='orange'
 />
 ```
 
--- a/src/components/Molecules/PostcodeLookup/PostcodeLookup.js
+++ b/src/components/Molecules/PostcodeLookup/PostcodeLookup.js
@@ -7,7 +7,7 @@ import { addressToString, addressFetcher } from './utils/PostcodeFunctions';
 import spacing from '../../../theme/shared/spacing';
 
 // ${({ theme }) => theme.linkStyles('standard')};
-const Button = styled.button`
+const AddressManually = styled.button`
   margin: ${spacing('md')} 0 0 ${spacing('sm')};
   background: inherit;
   border: none;
@@ -24,7 +24,8 @@ export default function PostcodeLookup({
   buttonText,
   noResultsMessage,
   reportError,
-  dropdownInstruction
+  dropdownInstruction,
+  buttonColour
 }) {
   const [showFields, setShowFields] = useState(false);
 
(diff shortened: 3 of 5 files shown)


src/components/Molecules/PostcodeLookup/PostcodeLookup.js
```diff
@@ -0,0 +1,70 @@
...
+
+const addressFetcher = async postcode => {
+  const valid = validatePostcode(postcode);
+  if (!valid) {
+    throw new Error('Please provide a valid postcode');
+  }
+  try {
+    return await getAddresses(valid);
+  } catch (error) {
+    // if (typeof Sentry !== 'undefined') {
```
AndyEPhipps, line 33: We shouldn't remove this, or we should at least provide a callback prop function that is called here, so we're able to run that Sentry function within the application this'll be used within.

src/components/Molecules/PostcodeLookup/PostcodeLookup.test.js
```diff
@@ -0,0 +1,11 @@
+import React from 'react';
+import 'jest-styled-components';
+
+import renderWithTheme from '../../../hoc/shallowWithTheme';
+import PostcodeLookup from './PostcodeLookup';
+
+it('renders correctly', () => {
```
AndyEPhipps, line 7: Can you add a few basic tests using the PCLU too plz

AndyEPhipps (changes requested):

src/components/Molecules/PostcodeLookup/PostcodeLookup.js
```diff
@@ -0,0 +1,70 @@
...
+ * @param onSelect
+ * @param rest
+ * @returns {JSX.Element}
+ * @constructor
+ */
+const PostcodeLookup = ({ onSelect, ...rest }) => (
+  <StyledWrapper>
+    <Lookup
+      name="postcode_lookup"
+      label="Find address by postcode"
```
AndyEPhipps, line 54: Let's make label, placeholder, buttonText and noResultsMessage overridable props with these values as the defaults.

AndyEPhipps:
Looking good, man! If we could just stop this button from drifting into the void on larger screens, I think we'll be good:

Screenshot 2024-05-09 at 11 40 29

Just to zoom out for a second, are we thinking about this existing as a 'drop-in' to be used to populate _already existing_ address fields within the parent form itself.. rather than it working like the [Storybook equivalent?](https://storybook.services.comicrelief.com/?path=/story/postcode-lookup--postcodelookup)

AndyEPhipps:
> @AndyEPhipps Not sure what you mean about the drop in usage, I think the existing storybook one is dropped into the current contact-us repo: https://github.com/comicrelief/contact-us/blob/master/src/components/Giro/Giro.js This would be used similarly. Probably not understanding what you are getting at sorry.

So I really meant the fact that this Lookup component doesn't include any _address_ fields itself; it just uses the Lookup API, then spits out a string made up of each part of the address.

How would you integrate this component within a real form?

AndyEPhipps (changes requested):
Aw jeez, dude, I feel like I wasn't being clear enough; I was really just _asking_ a legit question as to what was the plan of action, it wasn't a rhetorical one implying "we _should_ have address fields".

That said, this all looks nice, if we could just have functional parity with the [OG one here](https://storybook.services.comicrelief.com/?path=/story/postcode-lookup--postcodelookup); ('enter manually' button, show/hiding the fields, props to set button and error copy), we'll be onto a winner.




There'd definitely be benefits to

src/components/Atoms/Input/Input.js
AndyEPhipps, line 1: Hmm, all of these Input changes aren't right; it's stepping back updates we've made since this PR was started (phasing out `spacing`, adding `fontHelper`, using transient prop names etc.) - can you rework this please?

src/components/Atoms/TextInputWithDropdown/TextInputWithDropdown.style.js
```diff
@@ -19,7 +19,6 @@ const Dropdown = styled.div`
   border: ${({ $hideBorder }) => ($hideBorder ? 'none' : '1px solid')};
   margin-top: -1px;
   width: 100%;
-
```
AndyEPhipps, line 22: Can revert this too, no point in making stuff _less_ readable by bunching it all up together?

src/components/Molecules/Lookup/Lookup.js
```diff
@@ -1,30 +1,38 @@
 import React, { useCallback, useState } from 'react';
 import PropTypes from 'prop-types';
+import styled, { css } from 'styled-components';
 import TextInputWithDropdown from '../../Atoms/TextInputWithDropdown/TextInputWithDropdown';
 import spacing from '../../../theme/shared/spacing';
 import ButtonWithStates from '../../Atoms/ButtonWithStates/ButtonWithStates';
 
+const Button = styled(ButtonWithStates)`${({
```
AndyEPhipps, line 8: Given the size of this file, let's farm this out to its own style.js file

src/components/Molecules/Lookup/Lookup.js
```diff
@@ -77,7 +89,6 @@ const Lookup = ({
         options={options.map(mapOptionToString)}
         label={label}
         placeholder={placeholder}
-        data-test="lookupDropdown"
```
AndyEPhipps, line 80: Was gonna say "why are we removing this", but I can't see it used anywhere in our repos so no biggie I guess

src/components/Molecules/Lookup/Lookup.md
AndyEPhipps, line 1: I guess we can bin this file off now, since the new CL build doesn't use it?

src/components/Molecules/PostcodeLookup/utils/PostcodeFunctions.js
```diff
@@ -0,0 +1,53 @@
+import { isValid, toNormalised } from 'postcode';
+
+const validatePostcode = postcode => {
+  const trimmed = typeof postcode === 'string' ? postcode.trim() : '';
+  return isValid(trimmed) && toNormalised(trimmed);
+};
+
+// report Error - Sentry function passed as a prop
+const getAddresses = async (postcode, reportError) => {
+  const url = `https://lookups-staging.sls.comicrelief.com/postcode/lookup?query=${postcode}`;
```
AndyEPhipps, line 10: The endpoint should be set via a prop, so we can set it via an env var to fit the context this is used in.

src/components/Molecules/PostcodeLookup/PostcodeLookup.js
```diff
@@ -0,0 +1,104 @@
+import React, { useState } from 'react';
+import PropTypes from 'prop-types';
+import styled from 'styled-components';
+import Lookup from '../Lookup/Lookup';
+import AddressInputs from './utils/AddressInputs';
+import { addressToString, addressFetcher } from './utils/PostcodeFunctions';
+import spacing from '../../../theme/shared/spacing';
+
+const AddressManually = styled.button`
```
AndyEPhipps, line 9: Let's bung this in its own file too; it's hard to go back!

src/components/Molecules/PostcodeLookup/utils/AddressInputs.js
```diff
@@ -0,0 +1,90 @@
...
+  );
+}
+
+AddressInputs.propTypes = {
+  addressFields: PropTypes.shape({
+    postcode: PropTypes.string,
+    line1: PropTypes.string,
+    line2: PropTypes.string,
+    line3: PropTypes.string,
+    posttown: PropTypes.string
```
AndyEPhipps, line 88: Why not just 'town', as per all the previous versions?

src/components/Molecules/PostcodeLookup/utils/PostcodeFunctions.js
```diff
@@ -0,0 +1,53 @@
...
+    headers: {
+      'Content-Type': 'application/json'
+    },
+    timeout: 10000
+  };
+
+  try {
+    const response = await fetch(url, options);
+    if (!response.ok) {
+      throw new Error('Network response was not ok');
```
AndyEPhipps, line 22: Let's beef this up a bit by logging the error:

`throw new Error(JSON.stringify({ code: response.status, message: response.statusText }))` or something

src/components/Molecules/PostcodeLookup/PostcodeLookup.js
```diff
@@ -0,0 +1,104 @@
...
+        name={name}
+        label={label}
+        placeholder={placeholder}
+        buttonText={buttonText}
+        mapOptionToString={addressToString}
+        lookupHandler={postcode => addressFetcher(postcode, reportError)}
+        onSelect={handleAddressSelect}
+        dropdownInstruction={dropdownInstruction}
+        noResultsMessage={noResultsMessage}
+        buttonBgColour={buttonColour}
```
AndyEPhipps, line 67: Isn't there value in having all 4 of these actually be set-able via props?

src/components/Molecules/PostcodeLookup/PostcodeLookup.js
```diff
@@ -0,0 +1,104 @@
...
+}
+
+PostcodeLookup.propTypes = {
+  name: PropTypes.string,
+  label: PropTypes.string,
+  placeholder: PropTypes.string,
+  buttonText: PropTypes.string,
+  noResultsMessage: PropTypes.string,
+  dropdownInstruction: PropTypes.string,
+  reportError: PropTypes.oneOfType([
```
AndyEPhipps, line 89: Can just be `reportError: PropTypes.func`, no?

src/components/Molecules/PostcodeLookup/PostcodeLookup.test.js
AndyEPhipps: Given that there's easily-tested inputs and outputs (rather than like, trying to interact with a carousel or something), I think there's still value in adding this?

src/components/Molecules/PostcodeLookup/PostcodeLookupExample.jsx
```diff
@@ -0,0 +1,24 @@
...
+  }
+
+  return (
+    <ExampleContainer>
+      <PostcodeLookup
+        label="Postal Address"
+        buttonText="Find Address"
+        noResultsMessage="Sorry, we could not find your address."
+        dropdownInstruction="Dropdown instruction here..."
+        buttonColour="#f04257"
```
AndyEPhipps, line 19: We really wanna be using our colour variables for this example and within the associated code too

AndyEPhipps (commented):
Still needs a bit of chin-scratching I think, covered in the comments.

In addition, given the differences between our current piecemeal PCLUs used in various places, I think it's worth looping Curtis in here now, specifically looking at the dropdown for now:

The version used in the FSU currently is absolutely positioned below the postcode field, much like you've done here this new component:

Image

But Donate properly sticks it in the DOM flow, pushing the other fields down:

Image

Since the end goal here is to avoid repetition and endlessly remaking of the wheel, I think it's worth seeing if Curtis is happy with having all contexts align, and if not, extending the props to cover the customisations.

Turbo minor stuff:

Image
Lets bung an underline on this (to make it clearer it's a CTA, as per the previous versions) and sort the alignment with the button above.

PR #612, opened 2023-12-05
feat: 'Monthly only' widget copy fields

AndyEPhipps:
#### What is it doing?
Provides new props and logic to allow Content team to extend the usefulness of the Donate row but adding overrides for Title, Subtitle, Other Amount and Choose Amount copyfields, only applied to the Monthly giving type, when supplied. 

#### Why is this required?
"Donate widget not being as optimised for different audiences (single and monthly)"

#### link to Jira ticket:
https://comicrelief.atlassian.net/browse/ENG-2933

### Quick Checklist:


AndyEPhipps:
Here's a pre-existing version with no new fields, working as it always has:
https://deploy-preview-612--cr-component-library.netlify.app/#!/Donate/1


https://deploy-preview-612--cr-component-library.netlify.app/#!/Donate/39
The example above provides some more-obvious `otherAmountText` and `chooseAmountText` overrides to show the preexisting functionality, overriding the default props the component already had. Switching to Monthly shows the values for the monthly-only equivalents that are new for this PR.


The same in this example, but for the Title and Subtitle fields
https://deploy-preview-612--cr-component-library.netlify.app/#!/Donate/41

And all four:
https://deploy-preview-612--cr-component-library.netlify.app/#!/Donate/43

src/components/Organisms/Donate/Donate.js
```diff
@@ -95,9 +115,8 @@ const Donate = ({
       <Wrapper formAlignRight={formAlignRight} aria-live="polite">
         <Header formAlignRight={formAlignRight}>
           <HeaderInner>
-            {Boolean(subtitle) && (
+            {showCopy && (
```
AndyEPhipps, line 118: Have moved these nested checks to one single bit of logic

KrupaPammi (approved):

PR #619, opened 2024-02-08
feat: Padding options for Donate row

AndyEPhipps:
#### What is it doing?
Adds a new prop to override and control padding.

#### Why is this required?
To make the NJAONS page look less barren

#### link to Jira ticket:
Not linking so it doesn't annoyingly mark it as 'done' once this is merged:
comicrelief.atlassian.net/browse/ENG-3091

### Quick Checklist:

### DIFF
--- a/src/utils/_utils.js
+++ b/src/utils/_utils.js
@@ -1,13 +1,13 @@
 const handlePadding = paddingOption => {
   switch (paddingOption) {
-    case 'upper-off':
+    case 'upper_off':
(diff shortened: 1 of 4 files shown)


src/utils/_utils.js
```diff
@@ -0,0 +1,18 @@
+const handlePadding = paddingOption => {
```
AndyEPhipps, line 1: Made this a handy reusable function as it'll be used in a few other places down the line

PR #627, opened 2024-03-11
fix: Added mobileColor prop in Text component

7777akhil:
PR description

**What is it doing?**
Adding a mobileColor prop in Text component

**Why is this required?**
Easiest way to handle this issue is that we just set up our Text component with an optional mobile colour prop

**link to Jira ticket:**
[ENG-2979]

### Quick Checklist:

[ENG-2979]: https://comicrelief.atlassian.net/browse/ENG-2979?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/components/Atoms/Text/Text.js
+++ b/src/components/Atoms/Text/Text.js
@@ -31,6 +31,11 @@ export const BaseText = styled.span`
           }
         `
     : null)};
+    ${({ mobileColor, theme }) => mobileColor && css`
+    @media (max-width: 1023px) {
+      color: ${theme.color(mobileColor)};
+    }
+  `};
   ${({ size, theme }) => (size === 'm'
     ? css`
       font-size: ${theme.fontSize('s')};
@@ -45,7 +50,7 @@ export const BaseText = styled.span`
  *  Weight is checked for existence to prevent overriding the tag's css
  */
 const Text = ({
-  tag, children, height, weight, family, ...rest
+  tag, children, height, weight, family, mobileColor, ...rest
 }) => (
   <BaseText
     {...rest}
@@ -53,6 +58,7 @@ const Text = ({
     height={height}
     weight={weight}
     family={family}
+    mobileColor={mobileColor}
   >
     {children}
   </BaseText>
@@ -77,7 +83,8 @@ Text.propTypes = {
     PropTypes.arrayOf(PropTypes.node),
     PropTypes.node,
     PropTypes.string
-  ])
+  ]),
+  mobileColor: PropTypes.string
 };
 
 Text.defaultProps = {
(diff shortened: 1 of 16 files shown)


src/components/Atoms/Text/Text.md
```diff
@@ -1,46 +1,86 @@
...
+    tag="h1"
+    family="Anton"
+    uppercase
+    weight="normal"
+    size="super"
+    mobileColor="blue"
+  >
+    <b style={{ width: "100px", display: "inline-block" }}>H1 </b>
+    <Text style={{ textTransform: "none" }} tag="span" size="s">
+      <b>family:</b> Anton <b>weight:</b> Regular <b>size:</b> 80px / 5rem /
```
AndyEPhipps, line 13: Let's add some copy here to these ace new examples to clearly state what's happening:

`<b>family:</b> Anton <b>weight:</b> Regular <b>size:</b> 80px / 5rem / mobileColour: green` etc.

src/components/Atoms/Text/Text.js
```diff
@@ -31,6 +31,11 @@ export const BaseText = styled.span`
           }
         `
     : null)};
+    ${({ mobileColor, theme }) => mobileColor && css`
+    @media (max-width: 1023px) {
```
AndyEPhipps, line 35: Check out some other components to see how we use variables within media queries: https://github.com/comicrelief/component-library/blob/master/src/components/Molecules/Box/Box.js#L12

Which makes use of: https://github.com/comicrelief/component-library/blob/dcf690c82974595e65c1bdb918144cc9df5c1db4/src/theme/shared/breakpoint.js

AndyEPhipps (changes requested):

src/components/Atoms/Text/Text.js
```diff
@@ -31,6 +31,11 @@ export const BaseText = styled.span`
           }
         `
     : null)};
+    ${({ mobileColor, theme }) => mobileColor && css`
+  @media ${theme.breakpoint('medium')} {
+    color: ${theme.color(mobileColor)};
```
AndyEPhipps, line 36: Be sure to check your work before pushing, dude; this now applies the mobile **_above_** this breakpoint, not below it.

AndyEPhipps (changes requested):
Neeeearly there

src/components/Atoms/Text/Text.js
AndyEPhipps: Much better!

In future, it might be best to think of our styles as mobile-first, so you'd apply any mobile stuff to the non-mediaquery part [here](https://github.com/comicrelief/component-library/blob/4de8fc339e28e01eb447b6a5d95b5da2627284f2/src/components/Atoms/Text/Text.js#L8) (handling if there is and isn't a supplied mobileColor prop), and then override it within a media query.

BUT this'll work just fine, good stuff :)

PR #637, opened 2024-04-03
feat: creates extra Cards options / functionality.

michael-odonovan:
e.g.
or

#### What is it doing?
Creates extra Cards options / functionality.

#### Why is this required?
Content creators request.

#### link to Jira ticket:
https://comicrelief.atlassian.net/browse/ENG-3227

### Quick Checklist:

### DIFF
--- a/src/components/Molecules/Card/Card.md
+++ b/src/components/Molecules/Card/Card.md
@@ -16,7 +16,51 @@ import Link from '../../Atoms/Link/Link';
 </Card>;
 ```
 
-### Card row option
+### Card - row layout on small breakpoint only
+```js
+const defaultData = require('../../../styleguide/data/data').defaultData;import Text from '../../Atoms/Text/Text';
+import Link from '../../Atoms/Link/Link';
+
+<Card
+  imageLow={defaultData.image}
+  images={defaultData.images}
+  backgroundColor="grey_light"
+  height="auto"
+  smallBreakpointRowLayout={true}
+  mediumBreakpointRowLayout={false}
+>
+  <Text tag="h3" color="purple" size="xl">
+    Title
+  </Text>
+  <Text tag="p">
+    Your donations make a very real difference to people in the UK and around the world, whether you donate through Red Nose Day, Sport Relief or to Comic Relief at any other time of year. Here are some of the many stories that show how people are changing their lives with your support.
+  </Text>
+</Card>;
+```
+
+### Card - row layout on medium breakpoint only
+```js
+const defaultData = require('../../../styleguide/data/data').defaultData;import Text from '../../Atoms/Text/Text';
+import Link from '../../Atoms/Link/Link';
+
+<Card
+  imageLow={defaultData.image}
+  images={defaultData.images}
+  backgroundColor="grey_light"
+  height="auto"
+  smallBreakpointRowLayout={false}
+  mediumBreakpointRowLayout={true}
+>
+  <Text tag="h3" color="purple" size="xl">
+    Title
+  </Text>
+  <Text tag="p">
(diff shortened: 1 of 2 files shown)


src/components/Molecules/Card/Card.js
```diff
@@ -1,27 +1,64 @@
...
+import styled, { css } from 'styled-components';
 import PropTypes from 'prop-types';
 import Picture from '../../Atoms/Picture/Picture';
 import spacing from '../../../theme/shared/spacing';
 
 const Container = styled.div`
-  display: flex;
   position: relative;
-  flex-direction: column;
+  display: flex;
```
AndyEPhipps, line 9: Just as part of housekeeping since now that the Card styles are much bigger, could we move them into their own `Card.styles.js` file plz?

src/components/Molecules/Card/Card.js
```diff
@@ -1,27 +1,64 @@
...
-  padding: ${spacing('l')};
   display: flex;
   flex-direction: column;
+  padding: ${spacing('l')};
+  /* Check for smallBreakpointLayout prop coming from the CMS, otherwise column view default */
+  ${({ smallBreakpointLayout }) => (smallBreakpointLayout === 'Row')
+    && css`
+      padding: ${spacing('sm')};
+      h1, h2, h3, h4, h5 {
+       margin-bottom: 0.5rem;
```
AndyEPhipps, line 42: Would be cool to use the `spacing.js` variables whenever possible rather than hardcoding here

src/components/Molecules/PartnerLink/PartnerLink.test.js
```diff
@@ -112,6 +116,19 @@ it('renders correctly', () => {
       background-color: #E52630;
     }
 
+    @media (min-width:740px) {
+
+    }
+
+    @media (min-width:1024px) {
```
AndyEPhipps, line 123: As per previous comment etccccc

src/theme/shared/breakpoints.js
```diff
@@ -0,0 +1,14 @@
+export const sizes = {
+  mobile: 0,
```
AndyEPhipps, line 2: Based on your Slack comment, I'm happy to have this file sit alongside the existing `breakpoint.js` file while we update the integration, but _would_ prefer a clearer name to differentiate; maybe `allBreakpoints` or something?

We should 100% put a ticket in for each repo to update all references, so we can remove the original breakpoint file as soon as possible... although I guess it's really just CRcom, and your new Contact Us repo that would be using it?

AndyEPhipps (changes requested):

AndyEPhipps (changes requested):

src/components/Molecules/Card/Card.md
```diff
@@ -17,9 +16,34 @@ import Link from '../../Atoms/Link/Link';
 </Card>;
 ```
 
+### Card row option
```
AndyEPhipps, line 19: Looking really good, mate; could we just add 2 more examples showing the other combinations?

```
smallBreakpointRowLayout={false}
mediumBreakpointRowLayout={true}
```
  
and 

```
smallBreakpointRowLayout={true}
mediumBreakpointRowLayout={false}
```
  
Whenever we're having to do tricksy overlapping matrices of styles, it's always good to see as many of the possible variations shown; the amount of times I've missed issues by not doing so has taught me the hard way 🥴

AndyEPhipps (commented):
Might be worth checking this with Curtis as the padding seems bit minimal to me?

src/components/Molecules/Card/Card.md
AndyEPhipps: Again, I'm sorry I didn't think of these earlier; having to REvisit code a 2nd or 3rd time is v annoying

AndyEPhipps (approved):

PR #639, opened 2024-04-10
feat: move from breakpoints to new allBreakpoints

michael-odonovan:
#### What is it doing?
Change to new breakpoints that make sense.

#### Why is this required?
Code quality.

#### link to Jira ticket:
https://comicrelief.atlassian.net/browse/ENG-3254

### Quick Checklist:

### DIFF
--- a/src/theme/shared/allBreakpoints.js
+++ b/src/theme/shared/allBreakpoints.js
@@ -1,9 +1,9 @@
 export const sizes = {
-  mobile: 0,
-  tablet: 740,
-  desktop: 1024,
-  nav: 1150,
-  large: 1440
+  S: 0,
+  M: 740,
+  L: 1024,
+  Nav: 1150,
+  XL: 1440
 };
 
 export default size => {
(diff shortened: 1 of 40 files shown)


src/theme/shared/allBreakpoints.js
```diff
@@ -0,0 +1,14 @@
+export const sizes = {
+  mobile: 0,
```
AndyEPhipps, line 2: Probably should have thought of this earlier, but I wonder if now would be a really good time to align this naming with what Curtis is using, in _either_ direction. I'll have a word with him and get back to you; it should be easy enough to just find-replace these if so

AndyEPhipps: Okay, to properly align with the DESIGN SYSTEM, I'm gonna suggest we go down this route:

```
export const sizes = {
  S: 0,
  M: 740,
  L: 1024,
  Nav: 1150,
  XL: 1440
};
```

Technically, Curtis uses 'SW' (SmallWeb), MW etc, but both `Nav` and the current `Large` are purely devvy stuff; the latter being used for fullwidth components.

As such, there's no `NavW` or `XLW` (both of which make me feel ill), so let's keep it cleaner as above.

AndyEPhipps: (Sorry bout the back-and-forthing :/ )

AndyEPhipps (approved):
Good stuff, thanks for taking care of this, mate; there's so much crap in these repos, it's great to start unfucking it all

PR #655, opened 2024-07-11
chore: Move dependencies to devDependencies (to reduce library weight)

zhibek:
e.g.
or

#### What is it doing?
Move `dependencies` that don't need to be published (i.e. supporting dependencies, those not referenced in the codebase) to `devDependencies`.

#### Why is this required?
Reducing the list of `dependencies` reduces the weight of the library, improving performance.

#### link to Jira ticket:

### Quick Checklist:

### DIFF
--- a/babel.config.js
+++ b/babel.config.js
@@ -3,7 +3,10 @@ module.exports = function (api) {
     api.cache(true);
 
     const presets = [ ["react-app", { "absoluteRuntime": false }] ];
-    const plugins = [ ["babel-plugin-styled-components"] ];
+    const plugins = [
+        ["babel-plugin-styled-components"],
+        ["import", { "libraryName": "lodash", "libraryDirectory": "", "camel2DashComponentName": false }]
+    ];
 
     return {
         presets,
--- a/package.json
+++ b/package.json
@@ -21,13 +21,8 @@
     "url": "https://github.com/comicrelief/component-library.git"
   },
   "dependencies": {
-    "@babel/cli": "^7.21.5",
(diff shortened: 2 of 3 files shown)


babel.config.js
```diff
@@ -3,7 +3,10 @@ module.exports = function (api) {
     api.cache(true);
 
     const presets = [ ["react-app", { "absoluteRuntime": false }] ];
-    const plugins = [ ["babel-plugin-styled-components"] ];
+    const plugins = [
+        ["babel-plugin-styled-components"],
+        ["import", { "libraryName": "lodash", "libraryDirectory": "", "camel2DashComponentName": false }]
```
AndyEPhipps, line 8: Just so I'm clear as to what this is doing, it's overriding a lodash setting directly, so it won't collide with an ESlint setting in react-payments, or ...?

AndyEPhipps (approved):
Good stuff here, @zhibek - I never even thought about how all those dev dependencies come along for the ride with the CL

babel.config.js
zhibek: Hey @AndyEPhipps. I installed a babel plugin [babel-plugin-import](https://www.npmjs.com/package/babel-plugin-import) - the config you highlighted is the config for that plugin, setting it to optimise our `lodash` imports so we only import the parts of the `lodash` library we actually use (so we import 3x functions, not all ~100 functions that lodash has).

AndyEPhipps: Ah I see, very nice!

PR #688, opened 2025-01-06
chore: Adds test ID to lookup button

curlyfriesplease:
#### What is it doing?
I'm adding tests to the Donate FE, and this button I can't target because it's within our CompLib. Adding here so it has a testID.

#### Why is this required?
For the benefit of adding tests

#### link to Jira ticket:
[DF-182]

[DF-182]: https://comicrelief.atlassian.net/browse/DF-182?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/components/Molecules/Lookup/Lookup.js
+++ b/src/components/Molecules/Lookup/Lookup.js
@@ -120,7 +120,7 @@ const Lookup = ({
         loading={isSearching}
         disabled={isSearching}
         loadingText="Searching"
-        data-test={`lookupButton-${buttonText}`}
+        data-test="lookupButton"
       >
         {buttonText}
       </StyledButton>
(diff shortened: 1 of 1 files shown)


src/components/Molecules/Lookup/Lookup.js
```diff
@@ -121,6 +120,7 @@ const Lookup = ({
         loading={isSearching}
         disabled={isSearching}
         loadingText="Searching"
+        data-test={`lookupButton-${buttonText}`}
```
AndyEPhipps, line 123: Wouldn't it be better to have something _stable_ here, rather than something dynamic?

curlyfriesplease: Possibly yes, I was wondering about a situation in which a page might have >1 implementations of Lookup. Unlikely scenario, but in such a case the buttons are likely to be labelled differently, and then could be targeted separately. 
I've got no problem keeping it simple though so I'll strip it out.

AndyEPhipps: I don't believe I've ever see us have multiple lookups on one page before, so I think we're safe 👍

AndyEPhipps (approved):

PR #701, opened 2025-01-30
feat: Allows Logos to be exported, and adds ability to customise title

curlyfriesplease:
#### What is it doing?
Allows 'Logos' to be used in the frontend as it's not currently an export, and also allows a custom 'title' prop

#### Why is this required?
So that the logo can be used across multiple headers in various repos

#### link to Jira ticket:
[DF-205]

[DF-205]: https://comicrelief.atlassian.net/browse/DF-205?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/components/Molecules/Logos/Logos.md
```diff
@@ -1,5 +1,5 @@
 # Comic Relief Logo
 
 ```js
-<Logos rotate={false} sizeSm="50px" sizeMd="60px" />
-```
\ No newline at end of file
+<Logos animateRotate sizeSm="50px" sizeMd="60px" />
```
AndyEPhipps, line 4: We've previously had logos rotated a bit (by default, not on hover), so would rather we don't make this PR destructive in that respect

AndyEPhipps: https://github.com/comicrelief/component-library/blob/60ff6f0e8ef67d4b870ea643259e8dc1067b7f4f/src/components/Atoms/Logo/Logo.js#L28

AndyEPhipps: Although I guess this is just an example, so doesn't matter, now I thiink about it for 10 seconds

curlyfriesplease: Yeah I wasn't sure about it, opted _for_ it just to help show off the possible functionality within the CL, I reckon if there is functionality that's not immediately obvious it's always good to enable it, for the purposes of demoing it in your CL showcase thingy

AndyEPhipps: Could always add a second example that covers _this_ functionality, leaving the previous one in place (with some descriptive titles above); makes for even-clearer explanation of what we have and how to use it.

curlyfriesplease: Good shout. I've got two examples in place instead:

image

AndyEPhipps: Lovely stuff

AndyEPhipps (approved):
Looking good man; just now a bit confused why we ever had Logo _and_ Logo**s** in the first place 🤔 

Looks like Logo**s** is specific to the header nav, and it _sometimes_ includes [more than one Logo](https://github.com/comicrelief/component-library/blob/f1bc8f81f699f0f15c2e9117b947886f4ad70462/src/components/Molecules/Logos/Logos.js#L31).

Move to rename Logos to LogosWithLink or something, to delineate?

AndyEPhipps:
I'll spin up a CRcom PR to bring in these changes, updating the usage accordingly

PR #710, opened 2025-02-28
feat: changes look of Input error messages as per Curtis design for Donate

michael-odonovan:
e.g.
or

#### What is it doing?
Changes look of Input error messages as per Curtis design for Donate

#### Why is this required?
Design update.

#### link to Jira ticket:
https://comicrelief.atlassian.net/browse/DF-232

### Quick Checklist:


michael-odonovan:
@AndyEPhipps 
Have checked the look of all this with Curtis:
Screenshot 2025-03-03 at 6 04 58 PM
Screenshot 2025-03-03 at 6 05 12 PM

AndyEPhipps (commented):
Looks good man, maybe just a bit more right-padding here to separate the icon from the copy?

Screenshot 2025-03-04 at 10 00 47

Could be tweak only on-error if it looks a bit lopsided when there's no error icon

src/components/Atoms/Input/Input.js
```diff
@@ -2,45 +2,61 @@ import React from 'react';
...
+      position: absolute;
+      top: 50%;
+      right: 0.5rem;
+      transform: translateY(-50%);
+      background: url(${alertIcon}) center/contain no-repeat;
+      --iconSize: 19px;
+      width: var(--iconSize);
+      height: var(--iconSize);
+      z-index: 3;
+      margin-left: 20px;
```
AndyEPhipps, line 53: This won't do anything as the pseudo element is absolutely positioned; can probs just remove it

src/components/Atoms/Input/Input.js
```diff
@@ -86,7 +102,7 @@ const Input = React.forwardRef(
       optional={optional}
       {...labelProps}
     >
-      <InputWrapper>
+      <InputWrapper error={Boolean(errorMsg)}>
```
AndyEPhipps, line 105: Could we keep the operator within the render function?

AndyEPhipps (approved):

PR #714, opened 2025-04-08
Allows input field to have max Rem widths set at desktop breakpoints

curlyfriesplease:
e.g.
or

#### What is it doing?

#### Why is this required?

#### link to Jira ticket:
[DF-313]

### Quick Checklist:

### DIFF
--- a/src/components/Atoms/Input/Input.js
+++ b/src/components/Atoms/Input/Input.js
@@ -22,10 +22,14 @@ const InputFieldContainer = styled.div`
   display: flex;
   justify-content: flex-end;
   align-items: center;
-
-  @media ${({ theme }) => theme.allBreakpoints('M')} {
-    max-width: 290px;
-  }
+  ${({ maxRemWidthMediumBreakpoint, maxRemWidthLargeBreakpoint, theme }) => css`
+    @media ${theme.allBreakpoints('M')} {
+      ${maxRemWidthMediumBreakpoint && `max-width: ${maxRemWidthMediumBreakpoint}rem;`}
+    }
+    @media ${theme.allBreakpoints('L')} {
+      ${maxRemWidthLargeBreakpoint && `max-width: ${maxRemWidthLargeBreakpoint}rem;`}
+    }
+  `}
 `;
 
 const InputField = styled.input`${({ theme, error, prefixLength }) => css`
@@ -48,10 +52,6 @@ const InputField = styled.input`${({ theme, error, prefixLength }) => css`
 
   :focus {
     border: 1px solid ${theme.color('grey_for_forms')};
-  }
-
-  @media ${theme.allBreakpoints('M')} {
-    max-width: 290px;
   }
 `}`;
 
@@ -95,6 +95,8 @@ const Input = React.forwardRef(
       labelProps = {},
       prefix = '',
       optional = null,
+      maxRemWidthMediumBreakpoint,
+      maxRemWidthLargeBreakpoint,
       ...rest
     },
     ref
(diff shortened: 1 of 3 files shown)


src/components/Atoms/Input/Input.js
```diff
@@ -22,10 +22,14 @@ const InputFieldContainer = styled.div`
   display: flex;
   justify-content: flex-end;
   align-items: center;
-
-  @media ${({ theme }) => theme.allBreakpoints('M')} {
-    max-width: 290px;
-  }
+  ${({ maxRemWidthMediumBreakpoint, maxRemWidthLargeBreakpoint, theme }) => css`
+    @media ${theme.allBreakpoints('M')} {
+      ${maxRemWidthMediumBreakpoint && `max-width: ${maxRemWidthMediumBreakpoint}rem;`}
```
AndyEPhipps, line 27: Hm, why did you switch to rems?

curlyfriesplease: To follow the best practice of using relative over absolute, and only resorting to px where there is a definable need to have a fixed value. Rem supremacy! 👑 
If you're keen to have px, I propose I add that so that either maxPx or maxRem can be given.

michael-odonovan (changes requested):
Hey Jon,
I think adding props for max widths is a cool solution, nice one.
However like Andy, I think maybe we should stick to px for the breakpoints, as when layouts get messy it is quite handy to have the current solid px reference points in place.
So maybe the props are just `maxWidthMediumBreakpoint` and `maxWidthLargeBreakpoint` and leave off the rem in this part?:
```
@media ${theme.allBreakpoints('M')} {
      ${maxWidthMediumBreakpoint && `max-width: ${maxWidthMediumBreakpoint};`}
```
Then if we go full Automatic for the Rems later, we have the option there anyway...

src/components/Atoms/Input/Input.js
AndyEPhipps: Sorry, I missed this last week: I _do_ admire your chutzpah but there's difficulty in diverging in a _single_ context when all the rest of the styles work with pixels. 

Admittedly it's not a world-ending situation _here_, but any such changes really need to be a coordinated effort, not least of all involving Curtis

curlyfriesplease: A very fair point, I take it and will switch to px now.

curlyfriesplease:
@michael-odonovan @AndyEPhipps 
Hi guys, just pushed a commit to change it to a px based prop (props are now named `maxPxWidthMediumBreakpoint` and `maxPxWidthLargeBreakpoint`). Hopefully this does the trick!

michael-odonovan:
@curlyfriesplease 
I think it would be really great to leave off the px - sometimes it is very handy to set them to 100% as sizing is based on the parent. So leave off the px and it's all good.

curlyfriesplease:
> @curlyfriesplease I think it would be really great to leave off the px - sometimes it is very handy to set them to 100% as sizing is based on the parent. So leave off the px and it's all good.

They're optional props and will only override if used, so simply not using them should allow the component to default to 100% width. I've included `px` in the prop name as a reminder of what unit the prop is expecting, without it we'd need to inspect it within the CL each time to remind ourselves what it expects. 
What I could do if you like, is add a prop that defines what unit we want - `MaxWidthUnit` could accept the strings of `px`, or `rem`, or `percent` and the two override props would make use of whatever you like there. Default to px if not stated.

michael-odonovan:
@curlyfriesplease 
Ok, cool, like a normal div it default to 100% which is cool. All good as is.

michael-odonovan (approved):

PR #717, opened 2025-05-19
fix: Adjust new error message styling in the Marketing Prefs section

michael-odonovan:
e.g.
or

#### What is it doing?

#### Why is this required?

#### link to Jira ticket:

### Quick Checklist:

### DIFF
--- a/src/components/Organisms/MarketingPreferencesDS/MarketingPreferencesDS.style.js
+++ b/src/components/Organisms/MarketingPreferencesDS/MarketingPreferencesDS.style.js
@@ -23,7 +23,7 @@ const OuterWrapper = styled.div`
     height: 0;
     overflow: hidden;
     z-index: -1;
-    content: 
+    content:
       url(${EmailIconWhite})
       url(${PhoneIconWhite})
       url(${PostIconWhite})
@@ -239,19 +239,9 @@ const ExtraInfo = styled.span`
 `;
 
 const MPTextInput = styled(TextInput)`
-  // Correct the ErrorMsg offset for this context
-  > div + span span {
-    margin-top: -20px;
-  }
-
-  > span {
-    margin-bottom: .5rem;
-  }
-
   input {
     border: 1px solid  ${({ theme }) => theme.color('black')};
     background-color: ${({ theme }) => theme.color('white')};
-    margin-bottom: 10px;
     @media ${({ theme }) => theme.allBreakpoints('M')} {
       max-width: none;
     }


AndyEPhipps:
Better for sure, but we still need something to sort the accessibility contrast issues; maybe something like this?

Screenshot 2025-05-19 at 13 11 32

Can you check with Curtis plz?

AndyEPhipps (changes requested):
As discussed

michael-odonovan:
Curtis design:
curtis-mp

src/components/Organisms/MarketingPreferencesDS/_MarketingPreferencesDS.js
```diff
@@ -61,13 +61,23 @@ const MarketingPreferencesDS = ({
 
   const customId = id ? `marketing-preferences--${id}` : 'marketing-preferences';
 
+  // Check for field-specific errors
+  const hasEmailError = Boolean(errors.mp_permissionEmail || errors.mp_email);
```
AndyEPhipps, line 65: Do we actually see errors on `mp_permissionEmail` et al? Aren't these just optional checkboxes? Not sure where an error would happen there?

src/components/Organisms/MarketingPreferencesDS/MarketingPreferencesDS.style.js
```diff
@@ -75,24 +75,24 @@ const AssociatedFieldsName = styled.span`
...
-    background-color: ${theme.color('grey_medium')};
-  }
+  border-radius: 10px;
+  border: 1px solid ${theme.color('grey')};;
 
   &.selected {
-    background-color: ${theme.color('blue_donate')};
+    background-color: ${isError ? theme.color('red') : theme.color('blue_donate')};
+    &:hover {
+      background-color: ${isError ? theme.color('blue_donate') : theme.color('blue_donate')};
```
AndyEPhipps, line 93: Seems like you don't need this check since the outcome is the same in both scenarios?

This is obviously Curtis' call but it feels weird to me that the red bg-colour goes away on hover when its _still_ erroring? Would probs speed up the transition either way; it current feels like a weird interaction to me in generally, esp given how massive the wrapper is

src/components/Organisms/MarketingPreferencesDS/MarketingPreferencesDS.style.js
```diff
@@ -75,24 +75,24 @@ const AssociatedFieldsName = styled.span`
...
-  border: 1px solid ${theme.color('grey')};;
-  border-radius: 10px;
   background-color: ${theme.color('grey_light')};
   transition: background-color 0.3s, color 0.3s;
-
-  &:hover {
-    background-color: ${theme.color('grey_medium')};
-  }
+  border-radius: 10px;
+  border: 1px solid ${theme.color('grey')};;
```
AndyEPhipps, line 88: Can probs sack off that double semicolon while we're here

src/components/Organisms/MarketingPreferencesDS/MarketingPreferencesDS.style.js
```diff
@@ -75,24 +75,24 @@ const AssociatedFieldsName = styled.span`
...
 
   &.selected {
-    background-color: ${theme.color('blue_donate')};
+    background-color: ${isError ? theme.color('red') : theme.color('blue_donate')};
+    &:hover {
+      background-color: ${isError ? theme.color('blue_donate') : theme.color('blue_donate')};
+      border-color: ${theme.color('grey_4')};;
```
AndyEPhipps, line 94: _Does 2 semi-colons add up to a whole one?_

_🤔💩🤔_

AndyEPhipps (changes requested):
As mentioned

michael-odonovan:
@AndyEPhipps 
Have checked with Curtis and the current build/error colours is what we want.

AndyEPhipps (approved):
Good stuff, man; that secondary red on error hover is much better

PR #729, opened 2025-07-04
refactor: remove srTheme and associated unused colour list.

michael-odonovan:
e.g.
or

#### What is it doing?

#### Why is this required?

#### link to Jira ticket:

### Quick Checklist:


AndyEPhipps:
Hmmm.. before we start pulling up the carpets, I think we're best off getting Curtis' steer here. 

Having the mechanism and set-up for a secondary theme is _still_ a useful thing, especially when SR is still a regular campaign with fluid requirements, plus the whole website project and lack/change of agency may require this flexibility still.

I know we've moved away from trying to make SR _pages_ their own "thing" (with tweaked colours and fonts), but I'd rather get some definite commitments before yanking stuff out. Obvo it _can_ all be added again, but just trying to save us too much fannying about.

Given that our CMS can _only_ use one colour list anyways, it feels like at least having the CR and SR themes use a single, combined list is a good start, and we can get on Curtis' ass about retiring the 10 different greys etc

PR #756, opened 2025-10-15
fix: Postcode validation within MarketingPrefs: Replace simple regex with more robust methods from the postcode library

curlyfriesplease:
#### What is it doing?
The basic regex currently in use is not sufficient. This replaces it with more robust autofix and validation methods from the postcode library. 
The same approach is being taken in CRcom here https://github.com/comicrelief/comicrelief-contentful/pull/1775

#### Why is this required?
We have had contact into Supporter Care, about users mistakenly putting 1 instead of I, or 0 instead of O, and not understanding why they can't submit.

#### link to Jira ticket:
[ENG-4308]

### Quick Checklist:

[ENG-4308]: https://comicrelief.atlassian.net/browse/ENG-4308?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/components/Organisms/MarketingPreferencesDS/_MarketingPrefsConfig.js
```diff
@@ -108,7 +114,7 @@ const buildValidationSchema = overrideOptions => {
     mp_postcode: yup.string().when('mp_permissionPost', {
       is: val => (!(mpValidationOptions.mp_permissionPost.disableOption)
       && mpValidationOptions.mp_permissionPost[val]),
-      then: schema => schema.required('Please enter your postcode').matches(/^[a-zA-Z]{1,2}\d[a-zA-Z\d]?\s*\d[a-zA-Z]{2}$/, 'Please enter a valid postcode')
+      then: schema => schema.required('Please enter your postcode').test('postcode-valid', 'Please enter a valid postcode', validatePostcode)
```
seb-cr: Thanks for this, makes a lot of sense to update both lookup components!

With Yup this logic will need to be done in two steps; the first is a `transform` which should apply `fixPostcode` to string values (and return anything else untouched), the second is the validity test which should be simply `postcode.isValid`. If we have only the test, Yup will accept the "fixed" value but not return it to us, and we'll get the original un-fixed value passed into the submission payload.

Something else to double-check, that we are actually using the return value from `mpValidationSchema.validate(...)`.

curlyfriesplease:
Going to close this. It's still of use, this unmerged work, but will be better suited in our grand new validation library.

curlyfriesplease:
Mere hours after closing this, Matt Hutson has chased this up and would like it added! Hah. 
Reopening PR. 
When we do our validation library we'll have to include this too probably.

AndyEPhipps:
> Mere hours after closing this, Matt Hutson has chased this up and would like it added! 

Is there a Slack/email chain for this anywhere?

curlyfriesplease:
> > Mere hours after closing this, Matt Hutson has chased this up and would like it added!
> 
> Is there a Slack/email chain for this anywhere?

There is not, not that I've been involved in at least, it's all been verbal in our planning meetings. 

@matthutson please could you link to a convo, or forward an email request, if you have such a thing with whoever it was you said you were talking to about it.

matthutson:
Here you go: 
[Comic Relief Digital Workshop - playback and wrap.pdf](https://github.com/user-attachments/files/28709906/Comic.Relief.Digital.Workshop.-.playback.and.wrap.pdf)

curlyfriesplease:
OK in that case it is definitely the general postcode lookup I need to change rather than the MP specifically. 
Bear with me

curlyfriesplease:
Reclosing this as it's not needed within MP; it's instead being implemented in [Donate here](https://github.com/comicrelief/react-payments/pull/352) and [FSU here](https://github.com/comicrelief/comicrelief-contentful/pull/1775)

PR #871, opened 2026-04-15
fix: Amend behaviour of ambient video control button

curlyfriesplease:
#### What is it doing?
Two amendments to the behaviour of play/pause button for ambient video, following Curtis' review:
1. Increase fadeout from 2s to 3s
2. Make the button appear on initial load, so the user knows it's there

#### Why is this required?
Design QA

#### link to Jira ticket:
[ENG-5006]

### Quick Checklist:

[ENG-5006]: https://comicrelief.atlassian.net/browse/ENG-5006?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/components/Atoms/AmbientVideo/AmbientVideo.style.js
+++ b/src/components/Atoms/AmbientVideo/AmbientVideo.style.js
@@ -78,7 +78,7 @@ const PlayPauseButton = styled.button`
   align-items: center;
   justify-content: center;
   opacity: 0;
-  transition: opacity 0.2s ease 2s;
+  transition: opacity 0.2s ease 3s;
   ${zIndex('high')};
 
   // Play icon is shifted to the right slightly as it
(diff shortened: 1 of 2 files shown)


AndyEPhipps (approved):
Very nice, swanky improvement

It does make me wish for a CSS property for "_currently visible in viewport_" to completely close the loop (so here, the button wouldn't be hidden prematurely when offscreen), but still seems to be in the realm of crummy JS for the moment 😮‍💨

curlyfriesplease:
> Very nice, swanky improvement
> 
> It does make me wish for a CSS property for "_currently visible in viewport_" to completely close the loop (so here, the button wouldn't be hidden prematurely when offscreen), but still seems to be in the realm of crummy JS for the moment 😮‍💨

I think that could be done in JS with the Intersection Observer. Never used it myself but saw a talk on it once and it looks like a handy thing
Not going down that rabbithole today unless there is further Design QA in that regard!

curlyfriesplease:
Gah. It won't publish likely because the auto-publish rule ignores certain semantic PR prefixes. 
Going to retitle this as "fix: etc" and try again

PR #875, opened 2026-04-20
fix: Stop StyledVideo from loading both video files

curlyfriesplease:
#### What is it doing?
Addresses an issue where mobile users were loading in both video files.
I believe it's because Gatsby compiles its files not knowing the viewport size, and went with the default desktop URL. Once rendered, react recalculates with the real viewport, and switches the video to mobile. This meant the user was downloading both files.

Now, the AmbientVideo component makes use of `<source>`. There are two in order and the browser will use the first one that's valid. 
This should mean that React won't need to recompute anything. The browser will know what to do correctly when it first loads.

I've given it a test on CRCom and it appears to fix the bug.

#### Why is this required?
Bugfix, optimisation.

#### link to Jira ticket:
[ENG-5050]

### Quick Checklist:

[ENG-5050]: https://comicrelief.atlassian.net/browse/ENG-5050?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/components/Atoms/AmbientVideo/AmbientVideo.js
+++ b/src/components/Atoms/AmbientVideo/AmbientVideo.js
@@ -20,6 +20,8 @@ import {
 // Normalise webpack module object ({ default }) or string to video URL
 const normaliseSrc = value => (typeof value === 'string' ? value : value?.default);
 
+const belowLBreakpointMediaString = `(max-width: ${breakpointValues.L - 1}px)`;
+
 const AmbientVideo = ({
   src,
   srcMobile,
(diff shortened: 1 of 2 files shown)


AndyEPhipps:
> Addresses an issue where mobile users were loading in both video files.
I believe it's because Gatsby compiles its files not knowing the viewport size, and went with the default desktop URL. Once rendered, react recalculates with the real viewport, and switches the video to mobile. This meant the user was downloading both files.

Nicely sussed-out, mate; I can indeed see that it was working as expected purely within the CL example [here](https://cr-component-library.netlify.app/#!/DoateBanner/13).

AndyEPhipps (approved):

PR #876, opened 2026-04-21
fix: Relax validation of names, to allow special characters, apostrophes and hyphens

curlyfriesplease:
#### What is it doing?
José
André
Müller
François
All people who cannot currently put their proper names into ESU. 
This change will relax the name regex to look for 'Letters' as defined by Unicode, rather than strictly A-Z only, which will allow characters with diacritics like those above. Also, it will allow names with apostrophes and hyphens, like J-Train and D'Snazzle. 

#### Why is this required?
We have been adopting a similar approach elsewhere, like with FSU, so this brings some consistency.

#### link to Jira ticket:
[ENG-5017]

### Quick Checklist:

[ENG-5017]: https://comicrelief.atlassian.net/browse/ENG-5017?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/components/Organisms/EmailSignUp/EmailSignUp.test.js
+++ b/src/components/Organisms/EmailSignUp/EmailSignUp.test.js
@@ -1,7 +1,7 @@
 import React from "react";
 import "jest-styled-components";
 import renderWithTheme from "../../../../tests/hoc/shallowWithTheme";
-import { EmailSignUp, validationSchema } from "./_EmailSignUp";
+import { EmailSignUp, buildEsuValidationSchema } from "./_EmailSignUp";
 import RichText from "../../Atoms/RichText/RichText";
 import { useForm, FormProvider } from "react-hook-form";
 import { yupResolver } from "@hookform/resolvers/yup";
@@ -9,7 +9,7 @@ import { yupResolver } from "@hookform/resolvers/yup";
 const DummyForm = () => {
   const formMethods = useForm({
     mode: "onBlur",
-    resolver: yupResolver(validationSchema),
+    resolver: yupResolver(buildEsuValidationSchema()),
   });
   const { handleSubmit } = formMethods;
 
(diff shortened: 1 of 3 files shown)


curlyfriesplease:
Tested and accented characters save correctly to the Contact Store's parquet 

image

AndyEPhipps (approved):
I guess it's not much 'wasted effort' here ahead of the centralised validation rules we discussed last week, given that'll take a wee bit time to formalise; nice one

AndyEPhipps:
Not sure if you wanted to keep tests for this purely within CRcom or not; I'm happy to do that given it's currently the only context we use this in.

src/components/Organisms/EmailSignUp/EmailSignUp.test.js
```diff
@@ -1,15 +1,15 @@
...
+import { EmailSignUp, buildEsuValidationSchema } from "./_EmailSignUp";
 import RichText from "../../Atoms/RichText/RichText";
 import { useForm, FormProvider } from "react-hook-form";
 import { yupResolver } from "@hookform/resolvers/yup";
 
 const DummyForm = () => {
   const formMethods = useForm({
     mode: "onBlur",
-    resolver: yupResolver(validationSchema),
+    resolver: yupResolver(buildEsuValidationSchema()),
```
curlyfriesplease, line 12: This wouldn't have been doing anything before - `validationSchema` is not exported from that file, so it would have been doing `yupResolver(undefined)` i.e. no validation rules

curlyfriesplease:
Added a simple schema validation test for it.

src/components/Organisms/EmailSignUp/EmailSignUp.test.js
AndyEPhipps: whoops

AndyEPhipps (approved):

PR #879, opened 2026-05-05
fix: Reduce Input field font size and padding

curlyfriesplease:
#### What is it doing?
Reduces our Input fields' font size from 1.25rem to 1rem, and reduces its padding to a consistent 1rem all the way around.

#### Why is this required?
Design review following the bug ticket below highlighting truncated text.

#### link to Jira ticket:
[ENG-5018]

### Quick Checklist:

[ENG-5018]: https://comicrelief.atlassian.net/browse/ENG-5018?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/components/Atoms/Input/Input.js
+++ b/src/components/Atoms/Input/Input.js
@@ -6,6 +6,7 @@ import alertIcon from './assets/error-alert-icon-red.svg';
 import Label from '../Label/Label';
 import ErrorText from '../ErrorText/ErrorText';
 import zIndex from '../../../theme/shared/zIndex';
+import fontHelper from '../../../theme/crTheme/fontHelper';
 
 // This seems to get a decent approximation of the necessary width (without resorting to measuring
 //  the element with JS)
@@ -13,7 +14,7 @@ const getPrefixWidth = prefixLength => `calc(1.5rem + (${prefixLength} * 0.5rem)
 
 const InputWrapper = styled.div`
   position: relative;
-  font-size: ${({ theme }) => theme.fontSize('s')};
+  ${({ theme }) => fontHelper(theme, 'span')}
 `;
 
 const InputFieldContainer = styled.div`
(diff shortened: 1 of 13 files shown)


src/components/Atoms/Input/Input.js
```diff
@@ -13,7 +13,7 @@ const getPrefixWidth = prefixLength => `calc(1.5rem + (${prefixLength} * 0.5rem)
 
 const InputWrapper = styled.div`
   position: relative;
-  font-size: ${({ theme }) => theme.fontSize('m')};
+  font-size: ${({ theme }) => theme.fontSize('s')};
```
AndyEPhipps: Hm, kinda feels like the perfect time to align this with the Curtis-deemed styles we laid out in [fontConfig](https://github.com/comicrelief/component-library/blob/a83e16518f5ac032910fef1a5d27c21fa1a95ec4/src/theme/crTheme/fontConfig.js#L176)

Are you able to check in with him to suss out which of these 2 is correct, given they both came from him?

curlyfriesplease: Very good idea! Handy thing that fontHelper.
We've gone with the `span` tag. 1rem below L, and a comfy 1.25rem above. 
Checked with him and it looks good. Update pushed.

AndyEPhipps: But if these are global changes for all Inputs, we should be updating that `formFieldInput` config in fontConfig and using it, no?

curlyfriesplease: I had totally missed that being there! That would be more sensible to use wouldn't it. Amended.

AndyEPhipps (approved):
🫡

PR #893, opened 2026-06-10
fix: Monthly 'other amount' text fix for Donate Banner

curlyfriesplease:
#### What is it doing?
It was found that the label above the amount input field was never fully wired up. As part of the investigation it was found that we also have some redundant props, a hangover from the previous iteration of the Donate widget. This PR:
- Uses either the `otherAmountText` or `monthlyOtherAmountText` props to label the field
- Removes the two redundant props `chooseAmountText` and `monthlyChooseAmountText`. These used to sit above the moneybuy buttons, and we don't have that text anymore.

Once this package is available I'll do the CRcom PR to finish by removing the relevant fields from the CMS.

#### Why is this required?
Bugfix & tidyup

#### link to Jira ticket:
[ENG-5197]

### Quick Checklist:

[ENG-5197]: https://comicrelief.atlassian.net/browse/ENG-5197?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/components/Organisms/DonateBanner/Form/Form.js
+++ b/src/components/Organisms/DonateBanner/Form/Form.js
@@ -255,7 +255,7 @@ const Signup = ({
           )}
           <FormFieldset>
             <Label size="s" weight="500" color="black">
-              Enter another amount
+              {otherAmountText}
             </Label>
             <AmountField
               $hideMoneyBuys={hideMoneyBuys}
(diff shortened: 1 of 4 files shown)


src/components/Organisms/DonateBanner/Form/Form.js
```diff
@@ -313,12 +313,11 @@ Signup.propTypes = {
   cartID: PropTypes.string.isRequired,
   donateLink: PropTypes.string.isRequired,
   otherAmountText: PropTypes.string.isRequired,
+  monthlyOtherAmountText: PropTypes.string,
```
AndyEPhipps: This doesn't seem to get used, as the final copy is switched within `DonateBanner` via a`handleOtherAmountText` call based on the current givingType, and the single output is passed to the `Form` via `otherAmountText`

curlyfriesplease: Good spot

AndyEPhipps (approved):

PR #894, opened 2026-06-10
fix: Layout shift issue when accordion is opened

curlyfriesplease:
#### What is it doing?
Fixes a bug where, when there's a lot of text in an accordion, the layout of the text shifts during the opening animation. See the video in the attached Jira ticket.

The cause was some old CSS that is now useless, as well as being problematic.
When closed, at M breakpoint and above there is a 'lg' (3rem) lower padding on the text (even though we can't yet see the text!)
But when opened, this is then changed to 'l' (2rem). It makes this reduction at the same time as the opening animation, which results in the text visibly shifting.

Removing this CSS as it's redundant.
Also, amends an example to mimic our real world implementation where the problem was seen: it's fixed at the same width it is eventually implemented in, and has the same problematic text block currently in use on the prod site. 

#### Why is this required?
Bugfix

#### link to Jira ticket:
[ENG-5196]

### Quick Checklist:

[ENG-5196]: https://comicrelief.atlassian.net/browse/ENG-5196?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


AndyEPhipps (approved):

curlyfriesplease:
Will redo this off master as there's been significant changes since this was written

PR #898, opened 2026-07-16
fix: Import fonts more efficiently

curlyfriesplease:
#### What is it doing?
Our frontends can have their performance scores increased, if we change the way we’re importing fonts.

Currently the Component Library[ imports our Anton and Montserrat fonts, here in the global.css file. ](https://github.com/comicrelief/component-library/blob/master/src/theme/shared/global.css#L1)

When our frontends load, for best performance we ideally want fonts brought in early doors. 
With our current way of doing things: After the frontend has loaded its initial HTML + CSS, it’s then loading the JS bundle, with includes our CL, then reading that CSS within it, then making an import. The fonts end up coming in later than we want, and Lighthouse flags this as a performance hit.

We should

a) Still allow the CL to import fonts for dev purposes, but not as part of its exported package
b) Amend the frontends to import fonts themselves, in their index.html files

This PR amends the CL to remove the `@import` (fonts) command at the root of the problem. The CL will now import the fonts itself in the same way the frontends are going to - via their own `index.html` files.

#### Why is this required?
🚤  *SPEED*

#### link to Jira ticket:
[ENG-5226]

### Quick Checklist:

[ENG-5226]: https://comicrelief.atlassian.net/browse/ENG-5226?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/README.md
+++ b/README.md
@@ -28,6 +28,18 @@ $ yarn add @comicrelief/component-library
 import { ThemeProvider, crTheme } from '@comicrelief/component-library';
 ```
 
+#### Load fonts in your app's HTML
+We use Anton and Montserrat fonts. Add these to your frontend's `index.html` `<head>`:
+
+```
+<link rel="preconnect" href="https://fonts.googleapis.com" />
+<link rel="preconnect" href="https://fonts.gstatic.com" crossorigin />
+<link
+  rel="stylesheet"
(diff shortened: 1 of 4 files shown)


curlyfriesplease:
Once the package is available I will immediately followup with frontend implementations, as this is a breaking change.

AndyEPhipps (approved):
👌

PR #909, opened 2026-08-05
ci: Publish releases using OIDC instead of npm token

seb-cr:
#### What is it doing?

Replaces long-lived token authentication with OpenID Connect in the npm publishing workflow.

The workflow needs to be using npm ≥11.5.1 and Semantic Release ≥25.0.1. I've also configured GitHub Actions to be this package's trusted publisher on npm.

For more about how this works, see npm's [Trusted Publishing docs](https://docs.npmjs.com/trusted-publishers) and GitHub's [OpenID Connect docs](https://docs.github.com/en/actions/concepts/security/openid-connect) docs.

#### Why is this required?

As part of ongoing efforts to strengthen supply chain security, npm will remove the ability to publish using 2FA-bypass tokens from January 2027. Full details can be found in [this post](https://github.blog/changelog/2026-07-08-npm-install-time-security-and-gat-bypass2fa-deprecation/#2fa-bypass-tokens-will-no-longer-publish-directly).

Switching to Trusted Publishing allows us to continue to fully automate our publishing workflow.

#### link to Jira ticket:

[ENG-5182]

### Quick Checklist:

[ENG-5182]: https://comicrelief.atlassian.net/browse/ENG-5182

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -101,6 +101,11 @@ jobs:
     if: github.ref == 'refs/heads/master'
     needs: playwright
     runs-on: ubuntu-latest
+    permissions:
+      contents: write # to be able to publish a GitHub release
+      issues: write # to be able to comment on released issues
+      pull-requests: write # to be able to comment on released pull requests
+      id-token: write # to enable use of OIDC for trusted publishing and npm provenance
     steps:
       - name: Checkout
         uses: actions/checkout@v7
@@ -109,6 +114,9 @@ jobs:
         uses: actions/setup-node@v7
         with:
           node-version-file: '.nvmrc'
+
+      - name: Update npm # Trusted Publishing requires npm >=11.5.1
+        run: npm i -g npm@latest
 
       - name: Restore packages cache
         uses: actions/cache@v6
@@ -126,4 +134,3 @@ jobs:
         run: yarn semantic-release
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
-          NPM_TOKEN: ${{ secrets.NPM_PUBLISHING_TOKEN }}
--- a/package.json
+++ b/package.json
@@ -95,7 +95,7 @@
     "react": "^18.3.1",
     "react-dom": "^18.3.1",
     "react-test-renderer": "^18.3.1",
-    "semantic-release": "^17.4.6",
+    "semantic-release": "^25.0.9",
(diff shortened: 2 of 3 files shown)


seb-cr:
I thought I'd be able to test this before merging, but had forgotten that the non-master branch check takes place before anything else. So the build runs, but Semantic Release takes no action. Afraid this will have to be one of those "hope it works" PRs. For a confidence boost, the diff is practically identical to the corresponding Lambda Wrapper PR: https://github.com/comicrelief/lambda-wrapper/pull/1224

.github/workflows/main.yml
```diff
@@ -109,6 +114,9 @@ jobs:
         with:
           node-version-file: '.nvmrc'
 
+      - name: Update npm # Trusted Publishing requires npm >=11.5.1
+        run: npm i -g npm@latest
```
AndyEPhipps, line 118: I guess there's no "danger" is having this always be latest, rather than updating it to a specific >=11.5.1 version?

AndyEPhipps (approved):
Makes sense to me, thanks for the context-ful links!

(I couldn't seem to find the Trusted Publishers section though, as I was curious to have a looksie. Is access to this tied to a specific user? I'm `owner` already tho 😕  Couldn't see any such link when inside our Organisation on there either.

AndyEPhipps:
> (I couldn't seem to find the Trusted Publishers section though, as I was curious to have a looksie. Is access to this tied to a specific user? I'm `owner` already tho 😕 Couldn't see any such link when inside our Organisation on there either.

> _Navigate to your package settings on [npmjs.com](https://www.npmjs.com/) and find the "Trusted Publisher" section._

Ignore me, they meant _specific_ packages, not the 'Packages' screen 🤦

.github/workflows/main.yml
seb-cr: That's a good question, I wondered the same. My thinking was that latest ensures that we are using a version compatible with any new security changes that npm starts enforcing. You're right though, there is a risk if someone manages to publish a malicious (or even just accidentally broken) version of npm.

Once we upgrade to Node 26 we can remove this step completely, as the bundled version of npm will meet our needs.

seb-cr:
Looks good; we've got the all-important success message:

```
[8:30:36 AM] [semantic-release] [@semantic-release/npm] › ℹ  OIDC token exchange with the npm registry succeeded
```

PR #912, opened 2026-08-10
feat: Refactor carousels

AndyEPhipps:
e.g.
or

#### What is it doing?

#### Why is this required?

#### link to Jira ticket:

### Quick Checklist:


src/data/data.js
```diff
@@ -70,7 +70,7 @@ const carouselItemsComplete = {
   desktopHeight: 475,
   headerCopy: "Over the past two years, we've supported",
   peopleHelpedText: '11.7 million people',
-  autoPlay: false,
+  autoPlay: true,
```
AndyEPhipps, line 73: Seems like these data objects used in the examples file weren't configured right to match the labelling, doh.

